PR #7, opened 2018-12-14
Debug driver -- number-printing syscall driver for relocation debugging

jrvanwhy:
This adds a "debug" driver which only relies on the `command` syscall. This driver is for Rust userspace debugging; because relocations do not currently work, the standard libtock-rs console functionality is unreliable.

It enables an application to print 1, 2, or 3 unsigned integers using a single system call. It can be removed once Rust application support and libtock-rs console support are sufficiently reliable.

### DIFF
--- a/golf2/src/main.rs
+++ b/golf2/src/main.rs
@@ -33,7 +33,7 @@ pub mod digest;
 pub mod aes;
 pub mod dcrypto;
 pub mod dcrypto_test;
-pub mod debug;
+pub mod debug_syscall;
 
 use capsules::console;
 use capsules::virtual_uart::{UartDevice, UartMux};
@@ -73,7 +73,7 @@ pub struct Golf {
     aes: &'static aes::AesDriver<'static>,
     //rng: &'static capsules::rng::SimpleRng<'static, h1b::trng::Trng<'static>>,
     dcrypto: &'static dcrypto::DcryptoDriver<'static>,
-    debug: debug::Debug,
+    uint_printer: debug_syscall::UintPrinter,
 }
 
 static mut STRINGS: [StringDescriptor; 7] = [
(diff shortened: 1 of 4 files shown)


golf2/src/main.rs
```diff
@@ -345,6 +368,7 @@ impl Platform for Golf {
 //            capsules::rng::DRIVER_NUM   => f(Some(self.rng)),
             kernel::ipc::DRIVER_NUM       => f(Some(&self.ipc)),
             dcrypto::DRIVER_NUM           => f(Some(self.dcrypto)),
+            0x80000001                    => f(Some(&DEBUG_DRIVER)),
```
phil-levis-google: This should at the very least be a const defined in main.rs; better yet would be to factor this out into an actual separate debug driver module.

phil-levis-google (changes requested):

jrvanwhy:
It appears googlebot is confused. I've filed a bug with the relevant team.

phil-levis-google:
Cool. I made one suggestion for improving the code.

jrvanwhy:
I'll let you know when it's ready for further review.

golf2/src/main.rs
jrvanwhy: I split it out, and renamed it to just Debug. Let me know if you'd rather it be DebugDriver or something else; I renamed it to make it more consistent with the other drivers but I am concerned about name conflicts.

jrvanwhy:
This is ready for further review. I'll deal with the CLA stuff when review is done (it looks like googlebot will un-fix the CLA label every time I touch this PR).

phil-levis-google (commented):
If you're worried about name conflicts, then I'd suggest naming it for what it does, rather than what you're using it for. E.g., IntegerConsolePrint or something. Because, yeah, Debug is pretty overloaded.

phil-levis-google (approved):
I leave it up to you on name change.

phil-levis-google:
LGTM!

phil-levis-google:
Would you rather I merge and override the mistaken CLA gods?

jrvanwhy:
I can do the merge myself. Lets set the cla: yes label manually, rather than use GitHub's override functionality.

PR #10, opened 2019-01-29
Implement static linking

jrvanwhy:
Because ROPI-RWPI support in the compiler is incomplete, this changes the libtock-rs layout file to allow Rust apps to be statically-linked. With this change in place, all language features should work correctly.

The caveat is we lose the ability to dynamically load Tock applications and we need to predict the size of the TBF headers in the application image. When compiler support for ROPI-RWPI is complete I'll implement any necessary dynamic relocation steps and reverse this change.

This change depends on [Initialize .data and .bss in libtock-rs](https://github.com/tock/libtock-rs/pull/63).

### DIFF
--- a/golf2/apps/.cargo/config
+++ b/golf2/apps/.cargo/config
@@ -17,6 +17,6 @@ target = "thumbv7m-none-eabi"
 
 [target.thumbv7m-none-eabi]
 linker = "rust-lld"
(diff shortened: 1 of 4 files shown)


jrvanwhy:
Oops, this was supposed to merge libtock-rs:static-linking into libtock-rs not tock-on-titan:static-linking into tock-on-titan. Replaced by https://github.com/tock/libtock-rs/pull/64

PR #13, opened 2019-01-31
Add elf2tab as a submodule of tock-on-titan.

jrvanwhy:
This makes development of elf2tab changes much easier and will allow us to incorporate https://github.com/tock/elf2tab/pull/6 as soon as it is merged.


jrvanwhy:
This depends on #12 or #9

phil-levis-google (approved):
Given that we're going to increasingly depend on libtock-rs and libtock-rs is likely to require some further updates to elf2tab (as requirements change), proactively defining the dependency makes sense to me.

PR #14, opened 2019-01-31
Use libtock-rs's static linking support

jrvanwhy:
This makes Rust apps work correctly on golf2.


jrvanwhy:
This depends on #13 and https://github.com/tock/elf2tab/pull/6

phil-levis-google (approved):
Tests are good!

PR #15, opened 2019-01-31
Add a simple printing library.

jrvanwhy:
Advantages over core::fmt (already supported by libtock-rs):
  1. No dynamic polymorphism means it'll optimize better and generate less
     assembly (speeding up low-level debugging).
  1. Usage syntax is nicer than fmt::write (based on absl::StrCat):
       `print!("value: ", value, "\n");`
     rather than
       `fmt::write:(Console, format_args!("value: {}", value));`

I expect this to evolve over time. If it proves sufficiently valuable we can
upstream it into libtock-rs.

### DIFF
--- a/golf2/apps/h1b_tests/src/main.rs
+++ b/golf2/apps/h1b_tests/src/main.rs
@@ -16,18 +16,14 @@
 #![no_std]
 
 extern crate alloc;
+extern crate simple_print;
 extern crate tock;
 
 fn main() {
-    use tock::console::Console;
-
-    let mut console = Console::new();
+    use simple_print::{console,hex};
+    console!("Cat video count: ", 9001, "\nWhere we eat: ", hex(51966usize), "\n");
     loop {
-        use core::fmt::Write;
-        use tock::timer;
-        console
-            .write_str("Hello, World!\n")
-            .expect("Failed console write");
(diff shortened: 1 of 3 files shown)


jrvanwhy:
This depends on #14

phil-levis-google:
I'd suggest perhaps a different name than `print!`, mostly because the call is different from `print!` in the kernel. Or maybe we should change the kernel to be `kprint!`?

phil-levis-google (commented):
Looks great -- my one concern is that there's also a `print!` in the kernel, which has different syntax. We should avoid the name conflict. I sort of lean towards changing the kernel one to be `kprint!`.

jrvanwhy:
This is an alternative approach. I really want a printing solution that works with the Display/Debug traits for compatibility reasons (because the general Rust ecosystem uses those traits for printing), but Display/Debug are hardcoded to use dynamic polymorphism and therefore generate lots of code.

In situations where that's prohibitive -- for example, when I'm constantly digging through the generated assembly as I've been doing lately or when flash space is scarce -- we'll need an alternative.

This also has a conflict with the standard library's `print!` macro, so I'll rename it.

phil-levis-google (approved):

jrvanwhy:
Rename complete.

PR #16, opened 2019-01-31
Add a test harness, switching h1b_tests to a test executor.

jrvanwhy:
This allows us to run Rust unit tests directly on Tock as an application. In future changes, I'll add a tool to make it easy for unit tests to do assertions and set expectations (`assert!()` panics so it cannot be used in these unit tests).


phil-levis-google (changes requested):
This looks great! The one catch is it depends on `print!`; once we resolve that (please see #15 ), we should be good to go.

phil-levis-google (approved):
console! it is!

jrvanwhy:
> This looks great! The one catch is it depends on `print!`; once we resolve that (please see #15 ), we should be good to go.

Resolved.

PR #22, opened 2019-02-20
Usb transport

phil-levis-google:
This is a very large PR that incorporates full support for USB transport (via endpoint 1, EP1) for H1B as well as a U2F test application (userspace process in C) that passes 

https://u2fdemo.appspot.com/

There are two major limitations to the current code:
    1. There is not support for proper key derivation/use of the H1B key ladder. This means that while the device executes all of the U2F protocol mechanisms properly, it is insecure due to probably-very-insecure key derivation.
    1. There is no support for non-volatile storage: keys do not persist across power cycles. This means the code is useless from a practical standpoint, as you cannot have a persistent U2F identity.

The PR adds a new third-party directory, chromium-ec, for code and header files related to embedded Chromium.

To test this code, go to tock-on-titan/golf2 and run `make program`. This should install the U2F application (`golf2/apps/u2f_app`). You can then either run `U2FTest` or the U2F demo from the URL above. User input is mapped to userspace GPIO 1, which is pin 1 of port 0; a falling edge denotes user input. On gnubby2 test boards, this involves pushing the user button (one far from connector).


phil-levis-google:
One note -- this requires a small change to the Makefile in libtock-c (there was a bug), so it depends on this PR in libtock-c:

https://github.com/tock/libtock-c/pull/38

phil-levis-google:
Closing this PR to break it into multiple, smaller PRs.

PR #23, opened 2019-02-27
USB transport over EP1

phil-levis-google:
This pull requests adds support for USB transport over endpoint 1. It is the first of a series of pull requests that add kernel support and a userspace application for a U2F key.

This PR includes an application, u2f_test, which demonstrates being able to receive a frame over USB from a U2F test application. Compiling and installing golf2 will install this application. Then, run U2FTest to send a frame. You should see output such as:

```
Tock 1.0 booting. Initialization took 68070 tics.
Tock 1.0 booting. About to initialize USB.
Start main loop.
 
= Running U2F Transport Test =
1. Receiving a U2F frame over USB.
Set EP1 receive flags.
Handling endpoint 1 events: out true, in false
Out interrupts: 0x1
U2F: ep1 frame received.
U2F: get_slice
Set EP1 receive flags.
   Received
ff ff ff ff 86 00 08 5f 42 bb 9b 87 d5 b3 46 00 00 00 00 00 fe 7f 00 00 8a 43 4d 05 4a 7f 00 00 
22 00 00 00 00 00 00 00 80 1c b2 7d cf 55 00 00 70 9a d0 e1 fe 7f 00 00 d7 3b fd 7c cf 55 00 00 
1. Receiving a U2F frame over USB.
Set EP1 receive flags.
```
The significant changes are:
  - separate EP1 (transport) event handling from EP0 (control); lines 477-503 in `h1b/src/usb/mod.rs`;
  - add a lifetime to the USB structure so it can have a system call driver
  - adding a system call driver for the USB peripheral (`h1b/src/usb/driver.rs`)
  - adding a trait for USB transport `UsbHidU2f`

There's also a little bit of variable/enum renaming for clarity, and some source cleanup.


h1b/src/usb/mod.rs
```diff
@@ -61,7 +124,7 @@ enum USBState {
     WaitingForSetupPacket,   // Waiting for message from host
     DataStageIn,             // Sending data to host
     NoDataStage,             // Sending status (not data) to host,
-                             // e.g. in response to set command
+    // e.g. in response to set command
```
jrvanwhy: Line 127: was this an accidental change? It looks like you un-aligned a previously-aligned comment.

jrvanwhy (changes requested):

h1b/src/usb/mod.rs
phil-levis-google: Ooops -- fixed!

golf2/apps/u2f_test/Makefile
```diff
@@ -0,0 +1,34 @@
...
+TOCK_USERLAND_BASE_DIR = $(CURRENT_DIR)../../../third_party/libtock-c
+BUILDDIR ?= $(CURRENT_DIR)/build/$(TOCK_ARCH)
+
+C_SRCS   := $(wildcard *.c)
+
+OBJS += $(patsubst %.c,$(BUILDDIR)/%.o,$(C_SRCS))
+
+TOCK_APP_CONFIG = -Xlinker --defsym=STACK_SIZE=$$(STACK_SIZE)\
+                  -Xlinker --defsym=APP_HEAP_SIZE=$$(APP_HEAP_SIZE)\
+                   -Xlinker --defsym=KERNEL_HEAP_SIZE=$$(KERNEL_HEAP_SIZE)
```
jrvanwhy: nit: remove an extra space from line 27 (it is aligned one column over relative to the previous two lines)

golf2/apps/u2f_test/main.c
```diff
@@ -0,0 +1,56 @@
...
+    for (int i = 0; i < 64; i++) {
+      if (i % 32 == 0) {
+        printf("\n");
+      }
+      printf("%02x ", u2f_received_frame[i]);
+    }
+    printf("\n");
+    //printf("1. Transmitting a U2F error packet over transport.\n");
+    //ret = tock_u2f_transmit(u2f_command_frame, 64);
+    //printf("Return value: %i.\n", ret);
```
jrvanwhy: The stated description of this change is that it adds a U2F receive test, not a U2F send test. The commented out debug code -- and the u2f_command_frame array -- should probably be left out of this PR.

h1b/src/usb/mod.rs
```diff
@@ -403,8 +367,8 @@ impl USB {
         // Wait until reset flag is cleared or timeout
         while self.registers.reset.get() & (Reset::CSftRst as u32) == 1 &&
             timeout > 0 {
-            timeout -= 1;
-        }
+                timeout -= 1;
+            }
```
jrvanwhy, line 372: Would you mind reformatting this entire loop (lines 367-371)? It is confusing both before and after this change. Perhaps something like the following would be better:

```
for _ in 0..timeout {
    if self.registers.reset.get() & (Reset::CSftRst as u32) != 1 {
        continue;
    }
}
```

jrvanwhy (changes requested):

golf2/apps/u2f_test/Makefile
phil-levis-google: Fixed.

h1b/src/usb/mod.rs
phil-levis-google: Your code is cleaner, but it doesn't early exit the function if the first timeout fails. This is OK. Fixed.

golf2/apps/u2f_test/main.c
phil-levis-google: Yes. Those are holdovers from when I was manually constructing responses. Fixed.

h1b/src/usb/mod.rs
jrvanwhy: Ah yes my code is not equivalent. There's a few other ways to write the correct code instead -- e.g. I think the following is equivalent:

```
  while self.registers.reset.get() & (Reset::CSftRst as u32) == 1 {
    if timeout == 0 { return; }
    timeout -= 1;
  }
```

phil-levis-google: Yeah, let's go with this.

phil-levis-google:
OK, I've responded to all the comments. Ready for re-review.

jrvanwhy (approved):

PR #24, opened 2019-02-27
Third party code from chromium and elsewhere.

phil-levis-google:
This pull request adds support code copied from chromium-c needed for the U2F application whose licenses cause it to fall under third party code. It also adds a `Makefile.app` so this code can be compiled as an external library for a libtock-c application. Some of the files have minor changes to reflect being built within libtock-c rather than a standalone firmware image. The changes are summarized here:

  - `sha256.c` adds `tock_compat_sha256` so that direct calls to SHA256 can use the digest hardware accelerator. Incremental calls with SHA256_INIT and SHA256_UPDATE still use the software implementation. It also removes a few helper #defines which are no longer needed.
  - `include/sha256.h` adds a declaration for `tock_compat_sha256` as well as HMAC/SHA256 system calls that are implemented in hardware (through system calls).
  - `include/timer.h` removes the dependency on task IDs since Tock doesn't use/need them.
  - `include/common.h` removes the dependencies on `config.h` and `module_id.h` since Tock doesn't use/need them. 
  - `include/signed_header.h` removes the BUILD_ASSERT calls that verify the size of `struct SignedHeader`.


third_party/chromiumos-ec/Makefile.app
```diff
@@ -0,0 +1,27 @@
...
+CHROMIUM_OBJS   =  $(notdir $(patsubst %.c, %.o, $(CHROMIUM_CFILES)))
+
+define CHROMIUM_EC_RULES
+
+$$(BUILDDIR)/$(1)/chromiumos-ec:
+	$$(TRACEDIR)
+	$$(Q)mkdir -p $$@
+
+$(BUILDDIR)/$(1)/chromiumos-ec/sha256.o: $(CHROMIUMOS_DIR)/sha256.c | $$(BUILDDIR)/$(1)/chromiumos-ec
+#	echo $(BUILDDIR)
```
jrvanwhy: This file looks like it written for this PR (i.e. it's not pulled from chromium-ec). In that case, please remove the commented-out debug lines (14, 20, and 27)

third_party/chromiumos-ec/include/LICENSE
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2010 The Chromium OS Authors. All rights reserved.
```
jrvanwhy, line 1: Please move this file one level higher in the directory structure (i.e. move it into chromium-ec/)

jrvanwhy (changes requested):

third_party/chromiumos-ec/include/LICENSE
phil-levis-google: Is it OK that sha256.c (in the top level directory) has a different license?

third_party/chromiumos-ec/Makefile.app
phil-levis-google: Done.

third_party/chromiumos-ec/include/LICENSE
phil-levis-google: Done.

jrvanwhy: No, I guess we should leave it as-is.

phil-levis-google: Reverted.

jrvanwhy (approved):

PR #26, opened 2019-03-07
Crypto driver updates to support U2F.

phil-levis-google:
This updates the digest (SHA/HMAC) and AES drivers. The SHA driver has HMAC support added (needed for U2F application). The AES driver is made closer to the standard Tock API; the original implementation pre-dated Tock's HIL, so had defined its own; this revision removes key expansion and wiping secrets, since they are not part of `symmetric_encryption::Client` but were a part of the H1B-specific  `AESClient`.


h1b/src/crypto/sha.rs
```diff
@@ -80,10 +87,41 @@ impl DigestEngine for ShaEngine {
         Ok(())
     }
 
+    fn  initialize_hmac(&self, key: &[u8]) -> Result<(), DigestError> {
+        let ref regs = unsafe { &*self.regs }.sha;
+
+        regs.itop.set(0); // clear status
+        self.current_mode.set(Some(DigestMode::Sha256Hmac));
+
+        //regs.trig.set(ShaTrigMask::Stop as u32);
```
jrvanwhy: Remove?

jrvanwhy (approved):

h1b/src/crypto/sha.rs
phil-levis-google: Removed.

PR #28, opened 2019-03-09
Fix the rust code: update to match the current libtock-rs commit

jrvanwhy:
My libtock-rs update PR (#17 ) was never tested, and does not compile. This PR includes the fixes necessary to make h1b_tests build and run.

Unfortunately, as libtock-rs's fmt module was removed, the necessary changes include adding a new formatting library.


userspace/simple_fmt/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+mod tests {
+    use super::{Base,fmt_u32,i32_to_decimal};
+
+    // Runs the given callback in parallel for every value of a u32.
+    fn for_each_u32(mut callback: impl Fn(u32) + Copy + Send + Sync) {
+        // Rayon only supports parallel iteration for closed-open ranges, so we
+        // can't create a parallel iterator that iterates through the entire
+        // range. Instead we create an iterator that gets us almost all the way
+        // then calls the callback on the last value.
+        use rayon::iter::{IntoParallelIterator,ParallelIterator};
```
phil-levis-google, line 78: Is the intention that rayon is/will become pervasive in testing for libtock-rs/rust apps? If so, let's make that explicit. I'd hate each test to pull in its own particular flavor of testing support. Let's minimize dependencies. If this is the only place rayon will be used, I'd suggest cutting it; if it's the standard approach then let's keep it.

phil-levis-google (commented):
Just one comment on the rayon dependency.

phil-levis-google (commented):
Just the comment on rayon.

phil-levis-google (commented):
Just the comment on the rayon dependency.

userspace/simple_fmt/src/lib.rs
jrvanwhy: I do not expect to see rayon used for other tests.

Given our conversation today, we will be removing simple_print. As a result, I'm fine with deleting this test code. Would you prefer I delete the tests or leave the (temporary) dependency in place?

phil-levis: Either's fine with me -- whichever you'd prefer. I'm just wary of accumulating lots of very narrowly-used dependencies. It's also just a suggestion -- you can tell me you prefer otherwise.

phil-levis (approved):

userspace/simple_fmt/src/lib.rs
jrvanwhy: Thanks. Given that I'm removing this code soon I'll go ahead and commit as-is.

In the (near?) future we'll need to vendor our dependencies (except the Rust toolchain itself, hopefully), at which point pulling in a crate like Rayon will be much more difficult. That'll automatically make these decisions more thoroughly-thought-out.

PR #30, opened 2019-03-12
Add Common Makefile Targets documentation

jrvanwhy:


doc/make_targets.md
```diff
@@ -0,0 +1,47 @@
...
+  output. It should not operate recursively.
+* `test` runs any available tests. In the future, this may need to be split
+  between test types/mechanisms. For example, we may need tests that run on the
+  developer's machine (platform-independent) to be separate from tests that
+  require a dev board. This should operate recursively, but tests that require
+  deploying firmware images must run sequentially.
+
+Makefiles should implement the sensible subset of the above list. For example, a
+Makefile for a C application most likely should not implement `check` or `doc`,
+and a code library should not implement `program` or `run`. Like Bazel, trying
```
phil-levis-google, line 45: What's Bazel?

phil-levis-google (approved):

doc/make_targets.md
jrvanwhy: Google's build system: https://bazel.build/

Bazel is the open-sourced version -- its internal name is Blaze. It is the only build system allowed in google3 (the monorepo), so most Google employees will be familiar with it. When we need to make build-related design decisions and don't have a specific reason to do otherwise, I'd prefer to copy Cargo/Bazel's design decisions for consistency.

PR #32, opened 2019-03-13
Put the userspace Rust code into a Cargo workspace.

jrvanwhy:
This will increase re-use of compilation artifacts for Rust libraries, and removes the need to specify compilation options on a per-binary basis.


phil-levis-google (approved):
LGTM.

PR #33, opened 2019-03-13
C app Makefile refactor.

jrvanwhy:
Deduplicate the C app Makefiles, and add support for some of the common Makefile targets.

I will revise this alongside #30


phil-levis-google (approved):
Nice. This is great. Thank you.

PR #37, opened 2019-03-22
Remove the simple_print library

jrvanwhy:
Simple_print was supposed to generate smaller code, but because Rust language features (such as integer division) depend on core::fmt it doesn't actually reduce code size significantly.

Depends on #36


phil-levis-google (approved):
Bummer! But glad you measured this.

PR #41, opened 2019-03-28
Vendor xattr

jrvanwhy:
Depends on #40


phil-levis-google (approved):
Is this needed for CI? Is the CI system going to be in Rust? All good if so, just wanted to check. Or is this dependency from elf2tab?

jrvanwhy:
> Is this needed for CI? Is the CI system going to be in Rust? All good if so, just wanted to check. Or is this dependency from elf2tab?

The individual commit messages contain dependency graphs demonstrating why each vendored dependency is necessary. In this case, elf2tab depends on tar which depends on xattr.

In general, to make Cargo happy we need to give it a registry that contains all packages mentioned in elf2tab's Cargo.lock (with matching checksums!). It took me a while to understand *why* Cargo was so picky, but it's because Cargo.lock is how Cargo implements hermetic builds. Cargo promises that if you give it some source code, Cargo.toml, and a matching Cargo.lock, it will use *exactly* the dependencies specified in that Cargo.lock. Further, Cargo.lock is platform-independent and contains all transitive dependencies of elf2tab including optional dependencies.

PR #60, opened 2019-03-28
Vendor spin

jrvanwhy:
Depends on #59


phil-levis-google (approved):
Yuck spin locks! But I guess it's OK for host side stuff.

jrvanwhy:
> Yuck spin locks! But I guess it's OK for host side stuff.

Actually, spin is a dependency of libtock-rs's. libtock-rs currently uses linked-list-allocator, which does not assume a single-threaded environment and uses a spinlock to make its allocator threadsafe. Fortunately, it appears only a single spinlock is used, so the cost of the spinlock (wasted code size + ram) is small.

If we want to remove this requirement, we'll need to write our own Tock-specific allocator for libtock-rs.

PR #74, opened 2019-04-03
Update elf2tab

jrvanwhy:
Soon, I will add another Rust binary to the repository that will share its Cargo target directory with elf2tab. With a shared target/ directory we need to use the same toolchain. This elf2tab update changes the used toolchain from nightly-2018-04-19 to stable, allowing my future Rust binary to use the stable toolchain.

Depends on #73

### DIFF
--- a/third_party/Build.mk
+++ b/third_party/Build.mk
@@ -26,7 +26,7 @@ third_party/build: build/cargo-host/release/elf2tab
 .PHONY: third_party/check
 third_party/check:
 	cd third_party/elf2tab && \
-		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" cargo check -Z offline
+		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" cargo check
 
 .PHONY: third_party/devicetests
 third_party/devicetests:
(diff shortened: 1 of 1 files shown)


phil-levis-google (approved):
One comment -- your use of dependent pull requests makes it harder to review them. I.e., #74 includes all of the changes for #73. Although its changes are small, I had to click through a lot to do a custom compare to see this was the case, only really made possible by the fact that we have a small number of branches.

I see the tradeoff as "making it easier to see what's in this particular PR" versus "making sure that if this PR is merged it doesn't break things." I.e., if we merge #74 before #73 it'll still work because it includes everything in #73 . But the end procedural result is the same, which is that you can't merge #74 until #73 has been approved. 

Anyways, just a thought.

jrvanwhy:
> One comment -- your use of dependent pull requests makes it harder to review them. I.e., #74 includes all of the changes for #73. Although its changes are small, I had to click through a lot to do a custom compare to see this was the case, only really made possible by the fact that we have a small number of branches.
> 
> I see the tradeoff as "making it easier to see what's in this particular PR" versus "making sure that if this PR is merged it doesn't break things." I.e., if we merge #74 before #73 it'll still work because it includes everything in #73 . But the end procedural result is the same, which is that you can't merge #74 until #73 has been approved.
> 
> Anyways, just a thought.

Yeah, GitHub's review UI fails spectacularly here.

Most of my PRs consist of a single commit, so you should be able to review the commit on its own. To confirm that a later PR is a previous PR plus one commit, you can compare the hashes of the previous PR's commit. Does that make sense?

#73 was an exception to my "PRs consist of a single commit" rule, but I'm happy to make that rule 100% consistent if it helps you. With a 1-week review latency, I cannot practically work without producing dependent changes, but I'm happy to do what you need to make them as manageable as possible for you to review.

phil-levis-google:
Maybe just put a note in the PR message which commits are unique to this PR? Then it would be easy. I don't particularly care if there are multiple, although if there are more than a few then I'd probably use a compare. Doing compares occasionally is no big deal, especially if it's a complicated PR. I am just trying to figure out a way that the effort put into reviewing a PR is proportional to its complexity (i.e., no large minimum activation energy).

phil-levis-google:
And yes I think this interface is bad and promotes bad practices.

jrvanwhy:
> Maybe just put a note in the PR message which commits are unique to this PR? Then it would be easy. I don't particularly care if there are multiple, although if there are more than a few then I'd probably use a compare. Doing compares occasionally is no big deal, especially if it's a complicated PR. I am just trying to figure out a way that the effort put into reviewing a PR is proportional to its complexity (i.e., no large minimum activation energy).

Seems reasonable to me. I can't provide a commit hash because that may change if I update the PR, but I could give the first few words of the commit message (e.g. from GitHub's commit list for the PR). Would that be helpful? E.g.:

Commits unique to this PR:
* Update elf2tab.

phil-levis-google:
Yeah that would totally work. Plus seems easy on your end.

PR #82, opened 2019-05-17
U2F keyladder support

phil-levis-google:
This PR adds limited support for the H1B keyladder and has the U2F application invoke the keyladder to generate its keying material. This involves some changes to the digest interface to allow HMAC digests involving keyladder "certificates". The system call API is very narrow -- it allows a process invoke a step of the keyladder for a particular cert.

```
Bldr |53803887
retry|0
oops?|no
Himg =8D1E52A5..4B60EE60
Hfss =635D6FFB..6E85CEE3
Hinf =D0FC7426..BC208A3A
ump @44000
ock: booted in 68182 tics; initializing USB and loading processes.
Tock: starting main loop.
 
Startingtests.
Running test basic_test
Cat video count: 9001
What we eat: cafef00d
inished test basic_test. Result: ucceeded.
TEST_FINISHED: SUCCESS
```


golf2/src/digest.rs
```diff
@@ -53,9 +53,11 @@ impl<'a, E: DigestEngine + 'a> DigestDriver<'a, E> {
 
 impl<'a, E: DigestEngine> Driver for DigestDriver<'a, E> {
     fn command(&self, minor_num: usize, r2: usize, _r3: usize, caller_id: AppId) -> ReturnCode {
+        //print!("DigestEngine: command({}, {}, {})\n", minor_num, r2, _r3);
```
jrvanwhy: Line 56 appears to be a leftover debugging statement -- please remove.

golf2/src/digest.rs
```diff
@@ -113,39 +116,74 @@ impl<'a, E: DigestEngine> Driver for DigestDriver<'a, E> {
...
             },
             // Finalize hash and output to output buffer (arg: unused)
-            2 => {
+            3 => {
                 self.apps
                     .enter(caller_id, |app_data, _| {
                         match self.current_user.get() {
                             Some(cur) if cur.idx() == caller_id.idx() => {}
                             _ => {
+                                //debug!("Finalize busy, {}\n", caller_id.idx());
```
jrvanwhy: Another debugging statement that probably should be removed.

golf2/src/digest.rs
```diff
@@ -113,39 +116,74 @@ impl<'a, E: DigestEngine> Driver for DigestDriver<'a, E> {
...
                     })
                     .unwrap_or(ReturnCode::ENOMEM)
             },
+            4 => {
+                ReturnCode::SUCCESS
+                    /*if self.current_user.get().is_some() {
+                    ReturnCode::EBUSY
+                } else {
+                    ReturnCode::SUCCESS
+                }*/
```
jrvanwhy: Leftover code?

h1b/src/lib.rs
```diff
@@ -128,4 +128,6 @@ pub unsafe fn init() {
     cortexm3::nvic::disable_all();
     cortexm3::nvic::clear_all_pending();
     cortexm3::nvic::enable_all();
+
+    //cortexm3::nvic::Nvic::new(110).disable(); // SHA spins
```
jrvanwhy: Revert?

userspace/libgolf2/digest_syscalls.c
```diff
@@ -36,6 +51,11 @@ int tock_digest_hash_initialize(TockDigestMode mode) {
   return rval;
 }
 
+int tock_digest_cert_initialize(uint32_t cert) {
+  int rval = command(H1B_DRIVER_DIGEST, TOCK_DIGEST_CMD_CERT_INIT, cert, 0);
```
jrvanwhy: This can be a 1-liner, correct?

userspace/u2f_app/u2f_transport.c
```diff
@@ -407,13 +409,13 @@ static void u2fhid_cmd_init(U2FHID_FRAME *f_p) {
...
-          response.init.data[1], response.init.data[2], response.init.data[3],
-          response.init.data[4], response.init.data[5], response.init.data[6],
-          response.init.data[7]);
+  //printf("Response Frame -> cid:%08lx cmd:%02x ", response.cid,
+  //        response.init.cmd);
+  //printf("bcnth:%02x bcntl:%02x ", response.init.bcnth, response.init.bcntl);
+  //printf("data:%0X%0X%0X%0X%0X%0X%0X%0X\n\n", response.init.data[0],
+  //        response.init.data[1], response.init.data[2], response.init.data[3],
+  //        response.init.data[4], response.init.data[5], response.init.data[6],
+  //        response.init.data[7]);
```
jrvanwhy: I suspect these printf statements should be removed.

If we still want them, we should probably define a "debug_printf" or similar that is a no-op by default and can be enabled through a debug option (`make devprogram`?).

jrvanwhy (changes requested):
In general, we should probably have documentation for each system call driver. A .md file in docs/ with a table identifying the behavior of each minor_num would suffice.

userspace/u2f_app/u2f_transport.c
phil-levis-google: I agree that commented out printf statements are bad.

The issue I'm wrestling with is that when something goes wrong, you often want information from many different layers at the same time, but within a particular code path. I.e., you want debugging information about the control channel (EP0) across different parts of the stack, or you want data path information (EP1) across different parts of the stack. You can just reintroduce debug output each time, but this is a pain. So in the best case we have a set of debug flags you can turn on/off. You can see this in the three different debug macros within usb/mod.rs.

It's worth noting this particular printf() is from the original C code. :)

golf2/src/digest.rs
phil-levis-google: Done.

golf2/src/digest.rs
phil-levis-google: Done.

phil-levis-google: Should it be in doc/ or in libgolf2/ ?

jrvanwhy: I don't have a strong opinion on the topic. Keeping everything in doc/ makes it much more findable, with the risk that someone working with code forgets that the documentation exists. If you're putting documentation into libgolf2/ then it probably makes sense for that documentation to be in the form of a comment in the code.

jrvanwhy (approved):

jrvanwhy (approved):

userspace/u2f_app/u2f_transport.c
jrvanwhy: Yeah, this is a more general problem that we should solve at some point. A better debugging system would let us leave the debug messages in the code but make them no-ops (that compile away entirely) by default.

golf2/src/digest.rs
phil-levis-google: Put it in doc/.

h1b/src/lib.rs
phil-levis-google: Removed.

golf2/src/digest.rs
phil-levis-google: Ah, should have been commented back in; currently not used but needed in API.

userspace/libgolf2/digest_syscalls.c
phil-levis-google: Yes. Previously had debug output. Made into a one-liner.

userspace/u2f_app/u2f_transport.c
phil-levis-google: Just cut it.

PR #83, opened 2019-06-06
Cleanup of USB peripheral driver

phil-levis-google:
This is a partial code cleanup of the USB peripheral driver. It changes all of the registers that aren't direct integer values (i.e., have fields) to use the Register create. It also removes the remaining few magic constants inherited from the original code base and replaces them with properly formatted register fields/values.

I tested using U2FTest/u2f_app that the peripheral driver is still working properly.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
5b51cd0e78d4b4fae8a87f043c6ae85e2002bbce
git status
On branch usb_cleanup
Your branch is up to date with 'origin/usb_cleanup'.

```


h1b/src/usb/mod.rs
```diff
@@ -23,17 +23,30 @@ pub mod u2f;
...
+pub use self::registers::DeviceConfig;
+pub use self::registers::DeviceControl;
 pub use self::registers::DMADescriptor;
+pub use self::registers::EndpointControl;
+pub use self::registers::Gpio;
+pub use self::registers::InEndpointInterruptMask;
+pub use self::registers::Interrupt;
+pub use self::registers::OutEndpointInterruptMask;
+pub use self::registers::Reset;
+pub use self::registers::UsbConfiguration;
```
jrvanwhy: It may be worth doing `pub use self::registers::{AhbConfig, AllEndpointInterrupt, Descriptor, ...};` here.

h1b/src/usb/registers.rs
```diff
@@ -183,6 +458,9 @@ impl BitAnd for EpCtl {
     }
 }
 
+
+
+
```
jrvanwhy: Is this change intentional? 5 newlines in a row is a lot.

jrvanwhy (approved):

h1b/src/usb/mod.rs
phil-levis-google: Good point -- I've reformatted all of the use statements to be cleaner.

h1b/src/usb/registers.rs
phil-levis-google: Unintentional. Fixed.

PR #86, opened 2019-06-14
Enable read-write access to the last two flash pages.

jrvanwhy:
Longer term, we should probably have a nicer globalsec interface. This will suffice for a single global counter, as used by the U2F app.

Commits unique to this PR:

1. Enable read-write access to the last two flash pages via globalsec.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
d021a80a1f57a646257ff9a32f576ff30ac37690
git status
On branch flash-globalsec
```

### DIFF
--- a/golf2/src/main.rs
+++ b/golf2/src/main.rs
@@ -269,23 +269,57 @@ pub unsafe fn reset_handler() {
     // TODO(alevy): refactor out
     {
         use core::intrinsics::volatile_store as vs;
+        const GLOBALSEC_BASE:      usize = 0x40090000;
 
-        vs(0x40090000 as *mut u32, !0);
-        vs(0x40090004 as *mut u32, !0);
-        vs(0x40090008 as *mut u32, !0);
-        vs(0x4009000c as *mut u32, !0);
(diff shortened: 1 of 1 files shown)


golf2/src/main.rs
```diff
@@ -286,6 +286,24 @@ pub unsafe fn reset_handler() {
         vs(0x400900c4 as *mut u32, !0);
         vs(0x400900c8 as *mut u32, !0);
         vs(0x400900cc as *mut u32, !0);
```
phil-levis-google: Do you want to change these earlier globalsec operations to use your consts as well? It would be better if the two sets of accesses were consistent.

phil-levis-google (changes requested):

golf2/src/main.rs
jrvanwhy: Fine with me, done.

jrvanwhy:
Updated, ready for re-review.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
1914d0724d2d0318a2efda82c42b31084c559882
git status
On branch flash-globalsec
Your branch is up to date with 'origin/flash-globalsec'.

```

phil-levis-google (approved):

PR #89, opened 2019-06-14
Add the real flash hardware interface.

jrvanwhy:
Changes new in this PR:

1. Add the real flash hardware interface.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
8f2dc90a3a898df7d8e8d2b1940b1c887bcee116
git status
On branch real-flash
```

### DIFF
--- a/h1b/src/hil/flash/h1b_hw.rs
+++ b/h1b/src/hil/flash/h1b_hw.rs
@@ -1,0 +1,273 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use kernel::common::cells::VolatileCell;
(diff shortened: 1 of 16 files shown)


h1b/src/hil/flash/h1b_hw.rs
```diff
@@ -0,0 +1,249 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use kernel::common::cells::VolatileCell;
+
+#[repr(C)]
+pub struct H1bHw {
```
phil-levis-google, line 30: You should probably use the register crate here, e.g. as in 

https://github.com/google/tock-on-titan/blob/master/h1b/src/usb/registers.rs

It's probably fine to leave unused registers as VolatileCell<u32>s; entering lots of register definitions that aren't tested is more bug-prone than leaving unused registers for future people to enter and test.

phil-levis-google (changes requested):

h1b/src/hil/flash/h1b_hw.rs
jrvanwhy: I switched to the register crate where I think it made sense.

jrvanwhy:
Updated.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
31c7687a4990e2be2b10b5eafeb705c7e6b91046
git status
On branch real-flash
Your branch is up to date with 'origin/real-flash'.

```

phil-levis-google (approved):

PR #97, opened 2019-06-27
Add erase functionality to the flash driver.

jrvanwhy:
Commits new in this PR:

1. Add erase functionality to the flash driver.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
ec0b2ecc6bea55de9d0e7d700ebd83eb2c6c3608
git status
On branch flash-erase
Your branch is up to date with 'origin/flash-erase'.

```

### DIFF
--- a/userspace/h1b_tests/src/hil/flash/driver.rs
+++ b/userspace/h1b_tests/src/hil/flash/driver.rs
@@ -13,6 +13,9 @@
 // limitations under the License.
 
 #[cfg(test)]
+use { h1b::hil::flash::Hardware, kernel::hil::time::Alarm, test::require };
+
+#[cfg(test)]
 #[derive(Clone,Copy,PartialEq)]
 enum MockClientState {
 	EraseDone(kernel::ReturnCode),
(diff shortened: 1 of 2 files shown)


userspace/h1b_tests/src/hil/flash/driver.rs
```diff
@@ -44,17 +44,128 @@ impl h1b::hil::flash::Client for MockClient {
 	}
 }
 
+#[test]
+fn erase() -> bool {
+	use { h1b::hil::flash::Hardware, kernel::hil::time::Alarm, test::require };
```
phil-levis-google: I don't think I've seen `use` used in this way in Tock code before; we typically have the dependencies be explicit at the top of the file. A local `use` statement like this could be fine, I'm not sure. I just haven't seen it used like this and haven't thought through the implications. I'd suggest bringing it up on the call today? I can do it, if you'd like. The fact that it's in a test makes it less pressing.

phil-levis-google (approved):

userspace/h1b_tests/src/hil/flash/driver.rs
jrvanwhy: I think that "shortcuts" is a better description of `use` statements than "dependencies". I'd prefer to keep them in a narrower scope when possible, like any other declarations/definitions.

That said, there was some redundancy in the `use` statements so I moved the redundant ones to module scope.

jrvanwhy:
Updated

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
ad99ee24b968af7b4280080ec6eb3652c443021b
git status
On branch flash-erase
Your branch is up to date with 'origin/flash-erase'.

```

PR #99, opened 2019-07-12
USB suspend support

phil-levis-google:
This PR adds USB suspend/resume support to H1B. Sending a suspend from Linux causes the device to suspend properly. Resume is not seamless, due to some interactions with the Linux drivers. It takes a few messages to the device (or a few seconds) for Linux to fully recognize the device.

Note that this suspend/resume only operates on the USB peripheral: it does not currently put the core into a low-power state.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
ff0afbdcaad7d2f1bcf16771bfe19a94818693d3
git status
On branch usb_suspend_support
Your branch is up to date with 'origin/usb_suspend_support'.

```

### DIFF
--- a/golf2/src/main.rs
+++ b/golf2/src/main.rs
@@ -126,8 +126,8 @@ pub unsafe fn reset_handler() {
     let timerhs = {
         use h1b::pmu::*;
         use h1b::timeus::Timeus;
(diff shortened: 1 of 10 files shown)


phil-levis-google:
Looks like this introduces problems in initialization. Debugging now.

PR #103, opened 2019-08-07
Correct the `offset` value in the flash transaction parameters.

jrvanwhy:
The offsets are specified relative to the start of each flash module, whereas the `h1b::hil::flash::Hardware` interface uses offsets that cross over the two modules. H1bHw needs to do the mapping between them. Because only the second module is in use, we can simply hardcode the transformation.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
3ab4628b2ed60a01e391a60f9017608b901c443c
git status
On branch page-offset
Your branch is up to date with 'origin/page-offset'.

```

### DIFF
--- a/h1b/src/hil/flash/h1b_hw.rs
+++ b/h1b/src/hil/flash/h1b_hw.rs
@@ -261,6 +261,10 @@ impl super::hardware::Hardware<'static> for H1bHw {
 
 	fn set_transaction(&self, offset: usize, size: usize) {
 		use self::TransactionParameters::{Offset,Size};
+		// The offset is relative to the beginning of the flash module. There
+		// are 128 pages per flash module.
+		// TODO(jrvanwhy): Assumes the read is from the second flash bank.
+		let offset = offset - 128 * super::WORDS_PER_PAGE;
 		self.transaction_parameters.write(Offset.val(offset as u32) + Size.val(size as u32));
 	}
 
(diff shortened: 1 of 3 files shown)


h1b/src/hil/flash/h1b_hw.rs
```diff
@@ -261,7 +261,11 @@ impl super::hardware::Hardware<'static> for H1bHw {
 
 	fn set_transaction(&self, offset: usize, size: usize) {
 		use self::TransactionParameters::{Offset,Size};
-		self.transaction_parameters.write(Offset.val(offset as u32) + Size.val(size as u32));
+		// The offset is relative to the beginning of the flash module.
+		// TODO(jrvanwhy): Assumes the read is from the second flash bank.
+		self.transaction_parameters.write(
+			Offset.val(offset as u32 - 128 * 512)  // The first bank is 128 * 512 words
```
phil-levis-google: Would it make sense to define this as a const earlier in the file?

phil-levis-google (approved):

h1b/src/hil/flash/h1b_hw.rs
jrvanwhy: 512 appears in two places so I gave it its own constant. I added an explanation for the 128, as it does not appear elsewhere.

jrvanwhy:
```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
4ca1510ac4654f28023eada7ccb2d9cefe0d4d5f
git status
On branch page-offset
Your branch and 'origin/page-offset' have diverged,
and have 11 and 1 different commits each, respectively.

```

PR #104, opened 2019-08-07
Flash fixes

jrvanwhy:
1. Moved to timeout-based operation rather than interrupt-based
   operation due to b/138842998.
2. Add the final programming pulse needed for write operations.

Commits new in this PR:
1. Flash driver fixes. ...

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
0becf1dc0afba8bde1d8b8d8bc9ebe629bd72c03
git status
On branch flash-fixes
Your branch is up to date with 'origin/flash-fixes'.

```

### DIFF
--- a/h1b/src/hil/flash/driver.rs
+++ b/h1b/src/hil/flash/driver.rs
@@ -24,7 +24,7 @@ pub const WRITE_OPCODE: u32 = 0x27182818;
 /// The H1B flash driver. The hardware interface (either the real flash modules
 /// or the fake) is injected to support testing. This will not configure the
 /// globalsec flash regions -- that must be done independently.
-pub struct Flash<'d, A: Alarm + 'd, H: Hardware<'d> + 'd> {
+pub struct Flash<'d, A: Alarm + 'd, H: Hardware + 'd> {
     alarm: &'d A,
     client: Cell<Option<&'d Client>>,
 
@@ -46,7 +46,7 @@ pub trait Client {
 }
 
 // Public API for Flash.
-impl<'d, A: Alarm, H: Hardware<'d>> Flash<'d, A, H> {
+impl<'d, A: Alarm, H: Hardware> Flash<'d, A, H> {
     /// Constructs a driver for the given hardware interface. Unsafe because
     /// constructing multiple drivers for the same hardware seems like a bad
     /// idea. The caller must set the driver as the hardware's client before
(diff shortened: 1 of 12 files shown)


userspace/h1b_tests/src/hil/flash/smart_program.rs
```diff
@@ -48,23 +48,22 @@ fn successful_program() -> bool {
...
-	require!(alarm.get_alarm() == 150);
+	// Finish.
+	alarm.set_time(1_600_000);
+	hw.inject_result(0);
+	state = state.step(&alarm, &hw, WRITE_OPCODE);
+	require!(alarm.get_alarm() == 3_200_000);
 	require!(hw.is_programming() == true);
 	require!(state.return_code() == None);
-	alarm.set_time(100);
+	alarm.set_time(3_200_000);
```
phil-levis-google: Should this be set as a fraction of the frequency? Same question for similar cases below.

h1b/src/hil/flash/driver.rs
```diff
@@ -24,7 +24,7 @@ pub const WRITE_OPCODE: u32 = 0x27182818;
 /// The H1B flash driver. The hardware interface (either the real flash modules
 /// or the fake) is injected to support testing. This will not configure the
 /// globalsec flash regions -- that must be done independently.
-pub struct Flash<'d, A: Alarm + 'd, H: Hardware<'d> + 'd> {
+pub struct Flash<'d, A: Alarm + 'd, H: Hardware + 'd> {
```
phil-levis-google, line 27: Why is it parameterized by Alarm and Hardware? The standard approach in Tock is to only have the lifetime; if fields are references, then the structure doesn't require the actual types.

phil-levis-google (approved):

h1b/src/hil/flash/driver.rs
jrvanwhy: There are a few reasons I preferred this approach:

1. Using dynamic polymorphism instead of static polymorphism implies you need the dynamicism, and we don't.
2. It is probably cheaper (reduces the size of the driver by 1 word and allows inlining of method calls).

That said, I'm not strongly opinionated on the matter and if you still think I should switch to trait objects then I'll do so.

userspace/h1b_tests/src/hil/flash/smart_program.rs
jrvanwhy: These are now set in terms of MockAlarm's frequency. This is much more verbose -- do you still prefer it?

jrvanwhy:
```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
d2e2fcfecd622676009e6e6f657e243436da0512
git status
On branch flash-fixes
Your branch is up to date with 'origin/flash-fixes'.

```

jrvanwhy:
I'm going to submit this. Phil, if you have follow-up comments to my changes I'll take care of them in a second PR.

PR #108, opened 2019-09-06
Check bounds of offset on flash operations.

phil-levis-google:
This checks that the offset parameter passed to flash operations is within the bounds of the flash. Otherwise, a capsule can invoke a flash operation, pass a large offset, and read arbitrary RAM (allowing it to circumvent memory safety).

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
4f5aa012305f90f9121a1e85cc7fddbdcf15a689
git status
On branch flash_bounds_fix
Your branch is up to date with 'origin/flash_bounds_fix'.

```

### DIFF
--- a/h1b/src/hil/flash/h1b_hw.rs
+++ b/h1b/src/hil/flash/h1b_hw.rs
@@ -14,6 +14,7 @@
 
 #![allow(unused)]
 
+use kernel::ReturnCode;
 use kernel::common::cells::VolatileCell;
 use kernel::common::registers::ReadWrite;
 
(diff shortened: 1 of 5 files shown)


h1b/src/hil/flash/h1b_hw.rs
```diff
@@ -12,256 +12,274 @@
...
-		// TODO(jrvanwhy): Assumes the write is to the second flash bank (where
-		// the nvmem counter is).
-		self.pe_control_1.set(opcode);
-	}
+        fn is_programming(&self) -> bool {
+                // TODO(jrvanwhy): Only checks the second flash bank.
+                self.pe_control_1.get() != 0
+        }
+
+        fn read(&self, offset: usize) -> u32 {
```
jrvanwhy: Read should probably return a ReturnCode instead of silently reading 0.

jrvanwhy (changes requested):

h1b/src/hil/flash/h1b_hw.rs
phil-levis-google: I'm a bit worried about efficiency here. Returning a ReturnCode will require a two-word return (SuccessWithValue). For a low-level flash read that's really just a memory reference (e.g., code will do this in a tight loop), that might be a problematic overhead. It seems like it makes the common case (reading flash correctly) significantly slower for an edge case.

What do you think?

jrvanwhy: I don't think the overhead is significant, and should generally optimize away anyway. Another option is to have a read_with_default method instead, so callers pick the value rather than it being hardcoded to zero.

If you're still concerned about the overhead, though, let me know and I'll approve as-is.

phil-levis-google: It now returns a ReturnCode, with ReturnCode::ESIZE on an out-of-bounds offset.

jrvanwhy (approved):

PR #110, opened 2019-09-19
Add support for non-volatile storage of personality data

phil-levis-google:
This pull request adds support for the non-volatile storage of personality data. It has several parts:
    - Updates flash APIs so they can support virtualization (calls to `write` pass a buffer, which is returned back in `write_done`),
    - Updates flash write API to support arbitrary length writes (serialized into a sequence of 32 word writes) to avoid another almost-always-used layer of indirection/abstraction,
    - Adds a flash virtualization layer so the personality driver, non-volatile counter, and other flash abstractions can co-exist,
    - Modifies the globalsec settings so that the n-2 page (third from last) can be written, this is where personality data is stored,
    - Adds a system call driver for personality state,
    - Adds a `personality_test` application that checks if personality data is present, and if not generates it,
    - Adds a `personality_clear` application that clears personality data, and
    - Cuts 2 tests from h1b_tests in order to fit in application space.

------------------------------------------------------
```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
d9bb75316a16c6e2edc49589d5a16accec9a544d
git status
On branch h1b_personality_flash
Your branch is up to date with 'origin/h1b_personality_flash'.

```


h1b/src/hil/flash/driver.rs
```diff
@@ -65,18 +77,24 @@ impl<'d, A: Alarm, H: Hardware> super::flash::Flash<'d> for FlashImpl<'d, A, H>
         self.hw.read(word)
     }
 
-    fn write(&self, target: usize, data: &[u32]) -> ReturnCode {
-        if data.len() > 32 { return ReturnCode::ESIZE; }
-        if self.program_in_progress() { return ReturnCode::EBUSY; }
+    fn write(&self, target: usize, data: &'d mut [u32]) -> (ReturnCode, Option<&'d mut [u32]>) {
+        let write_len = cmp::min(data.len(), MAX_WRITE_SIZE);
+
+        //if data.len() > 32 { return (ReturnCode::ESIZE, Some(data)); }
```
jrvanwhy, line 83: We should delete this line instead of commenting it out, as the semantics of write have changed (and not temporarily).

I won't block this PR on this concern, though.

jrvanwhy (approved):

PR #122, opened 2019-10-03
Add a non-volatile counter capsule.

jrvanwhy:
This is a port of the nvcounter functionality from Cr50. Remaining tasks:

1. Add the capsule to the kernel
2. Write an integration test
3. Design and implement a syscall API

There are some changes to the test assertion library; these are not final and will be refactored in a future PR.

This depends on #121 . New commits in this PR:

1. Non-volatile counter capsule and its unit tests.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
61c9af76d8390375a65514fab1ec34e5dbcbbcd8
git status
On branch nvcounter
Your branch and 'origin/nvcounter' have diverged,
and have 1 and 1 different commits each, respectively.

```


h1b/src/nvcounter/traits.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// initialization or increment is ongoing. Note that callers must wait for
+    /// a Client::increment_done call to know whether the operation succeeded.
+    fn read_and_increment(&self) -> ReturnCode;
+
+    fn set_client(&self, client: &'c dyn Client);
+}
+
+/// Trait to be implemented by NvCounter clients.
+pub trait Client {
+    /// Called when a counter-initialization operation finishes.
```
phil-levis-google: Document what the possible ReturnCode values are and their meaning.

h1b/src/nvcounter/traits.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn initialize_done(&self, status: ReturnCode);
+
+    /// Called when an increment operation completes.
+    fn increment_done(&self, status: ReturnCode);
```
phil-levis-google, line 51: Document what the possible ReturnCode values are and their meaning. E.g., ReturnWithValue.

h1b/src/nvcounter/capsule.rs
```diff
@@ -0,0 +1,293 @@
+// Copyright 2019 Google LLC
```
phil-levis-google, line 1: Generally, capsule refers to a property of the code (a capsule has no unsafe code). But all of this code seems to be capsule code (e.g., the internal code too). I don't really have a better name, except maybe ::driver?

h1b/src/nvcounter/capsule.rs
```diff
@@ -0,0 +1,293 @@
...
+use super::traits::{Client,NvCounter};
+use crate::hil;
+
+/// NvCounter implementation using flash memory.
+
+// Uses two pages of flash, flipping two bits to zero in each page at a time.
+// The low page is struck or erased for every increment operation. When the low
+// page wraps around, the high page is struck once, the low page is erased, then
+// the high page is struck again. This allows the counter to resume if the
+// increment operation is interrupted, providing atomicity.
```
phil-levis-google, line 26: I suggest showing a calculation of how big the counter is (i.e., how many bits).

h1b/src/nvcounter/capsule.rs
```diff
@@ -0,0 +1,293 @@
...
+//
+// initialize() sequence:
+//   A1. Erase low
```
phil-levis-google: I'd suggest using names rather than symbols. E.g., call then sequence EraseLow, EraseHigh, WriteLow, WriteHigh, EraseLow, WriteHigh2. Referring to C2, A2, in the code is confusing.

h1b/src/nvcounter/capsule.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+}
+
+impl <'c, F: hil::flash::Flash<'c> + 'c> NvCounter<'c> for FlashCounter<'c, F> {
+    fn initialize(&self) -> ReturnCode {
+        // For now, we only support doing a single operation at a time.
+        if self.task.get().is_some() { return ReturnCode::EBUSY; }
+        // Try to start the erase (step A1). If a flash operation is ongoing, we
+        // will get back EBUSY. In that case, return success, as the erase will
```
phil-levis-google: Under what circumstances should EBUSY be returned from `erase` when there is no task? If there's broken virtualization? Or is the driver assuming no virtualization? If so, will it get callbacks (e.g., is it the client)?

phil-levis-google (approved):

h1b/src/nvcounter/capsule.rs
jrvanwhy: I don't have a better name either. ::driver would imply that this is a system call driver, and it isn't (that's a separate chunk of code).

h1b/src/nvcounter/capsule.rs
jrvanwhy: I added a calculation, currently lines 28-33.

h1b/src/nvcounter/traits.rs
jrvanwhy: Done.

h1b/src/nvcounter/traits.rs
jrvanwhy: Done.

h1b/src/nvcounter/capsule.rs
jrvanwhy: I updated them. I kept the (sequence, step number) pattern but gave the sequences more meaningful names.

h1b/src/nvcounter/capsule.rs
jrvanwhy: I added an explanation. EBUSY will be returned when a rollover increment is being finished in the background.

PR #125, opened 2019-10-22
Add the print_memory_usage.py script.

phil-levis-google:
This adds the Python3 script print_memory_usage.py, which takes a Tock ELF executable and prints out information about its flash and RAM use.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
69ae857713a45b02654da351e70af949bf467f64
git status
On branch space-script
Your branch is up to date with 'origin/space-script'.

```


tools/print_memory_usage.py
```diff
@@ -0,0 +1,386 @@
+#!/usr/bin/env python3
```
jrvanwhy, line 1: Reminder: Copyright header.

jrvanwhy (approved):
You created a tools/ directory, and I created a size_tools/ directory in #126 . Your script and my library probably belong in the same place. I decided to name the directory in #126 "size_tools" because it's more than just tool binaries (e.g. it contains a library); I think it's more sensible to rename tools/ to size_tools here than for me to move my library from size_tools to tools/.

What do you think?

tools/print_memory_usage.py
phil-levis-google: Yeah, just put that in. Realized it when I started reviewing your PR. Thanks.

phil-levis-google:
WRT tools vs. size_tools, I guess I see qualifiers as useful when the namespace becomes busy. These are both tools. Having a tools directory reflects what's in the main Tock repository. What happens if we start having more tools that aren't for size? Should each set have its own top level directory?  I'm fine with size_tools if you prefer it, but would myself lean towards tools/, with then a tools/size if tools becomes busy.

As for the presence of libraries, we could have tools/ and tools/libs.

jrvanwhy:
> WRT tools vs. size_tools, I guess I see qualifiers as useful when the namespace becomes busy. These are both tools. Having a tools directory reflects what's in the main Tock repository. What happens if we start having more tools that aren't for size? Should each set have its own top level directory? I'm fine with size_tools if you prefer it, but would myself lean towards tools/, with then a tools/size if tools becomes busy.
> 
> As for the presence of libraries, we could have tools/ and tools/libs.

Okay, seems reasonable to me. I'll switch my PR from size_tools/ to tools/.

PR #126, opened 2019-10-22
Implement the size_graph library.

jrvanwhy:
size_graph scans through an ELF file and forms a directed graph of symbols in the ELF binary and their sizes. It is intended to be used for the following types of analysis:
  1. What created the size delta between these two binaries?
  2. What large symbols have few references to them (i.e. where are
     the low-hanging size optimizations?)
  3. Why was a symbol pulled in?

I vendored the rustc-demangle library for Rust name demangling.

Work that remains to be done:
  1. Implementing binaries to perform the above analysis. The
     binary comparison library could benefit from a parallel loader,
     which I will implement.
  2. Looking into name mangling: Rust has a new and much more informative
     (but technically unstable) name mangling scheme. We may want to pull
     information out of the mangled names to make size analysis easier.
  3. Improve analysis of indirect calls -- this may be fine as-is but it
     may need work.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
27cf42c953350f676cd77d70131552c43ba9c0e2
git status
On branch size-tools
Your branch is up to date with 'origin/size-tools'.

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Subdirectories containing Build.mk files.
-BUILD_SUBDIRS := golf2 runner third_party userspace
+BUILD_SUBDIRS := golf2 runner third_party tools userspace
 
 .PHONY: all
 all: build
--- a/third_party/Build.mk
+++ b/third_party/Build.mk
@@ -26,7 +26,11 @@ third_party/build: build/cargo-host/release/elf2tab
 .PHONY: third_party/check
 third_party/check:
 	cd third_party/elf2tab && \
-		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" cargo check --frozen
+		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" \
+		cargo check --frozen --release
+	cd third_party/rustc-demangle && \
+		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" \
+		cargo check --offline --release
 
 .PHONY: third_party/devicetests
 third_party/devicetests:
@@ -34,12 +38,20 @@ third_party/devicetests:
 .PHONY: third_party/doc
 third_party/doc:
 	cd third_party/elf2tab && \
-		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" cargo doc --frozen
+		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" \
+		cargo doc --frozen --release
+	cd third_party/rustc-demangle && \
+		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" \
+		cargo doc --offline --release
 
 .PHONY: third_party/localtests
 third_party/localtests:
 	cd third_party/elf2tab && \
-		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" cargo test --frozen
+		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" \
+		cargo test --frozen --release
+	cd third_party/rustc-demangle && \
+		CARGO_BUILD_TARGET_DIR="../../build/cargo-host" \
+		cargo test --offline --release
 
 
 .PHONY: build/cargo-host/release/elf2tab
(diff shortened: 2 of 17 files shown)


phil-levis-google:
Can you fix the prtest?

jrvanwhy:
> Can you fix the prtest?

Would you mind clarifying? Seems fine to me.

phil-levis-google:
I read this as an issue, that what you're prtesting is not necessarily what's in the origin branch? I.e., you'll need to merge then prtest that merge? Or is this not a problem and I misunderstand?

```
Your branch and 'origin/size-tools' have diverged,
and have 1 and 1 different commits each, respectively.
```

jrvanwhy:
> I read this as an issue, that what you're prtesting is not necessarily what's in the origin branch? I.e., you'll need to merge then prtest that merge? Or is this not a problem and I misunderstand?
> 
> ```
> Your branch and 'origin/size-tools' have diverged,
> and have 1 and 1 different commits each, respectively.
>   (use "git pull" to merge the remote branch into yours)
> ```

Ah, that's because my workflow is:

1. git commit
2. make prtest
3. git push

You can check that the listed commit hash (74cb604...) matches the value that GitHub sees. As an alternative, I could push before I run prtest, which would give nicer prtest output, but then I need extra force pushes if the prtest fails.

jrvanwhy:
I've renamed size_tools to tools/, as discussed in #125 .

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
b69a91418ad2a135bb81d8d4c8cbd8d64ec14205
git status
On branch size-tools
Your branch is up to date with 'origin/size-tools'.

```

phil-levis-google:
I wanted to try the tool to see if I had any feedback on its use/output. When I run `make` in the `tools` directory, I get 

```
plevis@domrizzo4-w:~/src/jrvanwhy/tock-on-titan/tools$ make
make -C .. tools/build
make[1]: Entering directory '/usr/local/google/home/plevis/src/jrvanwhy/tock-on-titan'
cd tools && cargo build --offline --release
error: Found argument '--offline' which wasn't expected, or isn't valid in this context

USAGE:
    cargo build [OPTIONS]

For more information try --help
```
Do I have the wrong cargo version?

jrvanwhy:
> I wanted to try the tool to see if I had any feedback on its use/output.

There is no binary yet, just a library. I'm trying to separate out the gory/hacky implementation details (parsing objdump output, in this PR) from the UI design details ("coming soon").

> Do I have the wrong cargo version?

Yes. I was missing a rust-toolchain file, which I have now added. However, it just points at the latest stable toolchain, so you'll need to update your toolchain. This is a relatively-new cargo flag.

phil-levis-google:
> Yes. I was missing a rust-toolchain file, which I have now added. However, it just points at the latest stable toolchain, so you'll need to update your toolchain. This is a relatively-new cargo flag.

If you could include a check that this is properly updated, that would be good. E.g., I haven't updated to the latest stable toolchain (e.g., for elf2tab) for a long while, and figuring out that I needed to do so wasn't obvious.

phil-levis-google (approved):

jrvanwhy:
> > Yes. I was missing a rust-toolchain file, which I have now added. However, it just points at the latest stable toolchain, so you'll need to update your toolchain. This is a relatively-new cargo flag.
> 
> If you could include a check that this is properly updated, that would be good. E.g., I haven't updated to the latest stable toolchain (e.g., for elf2tab) for a long while, and figuring out that I needed to do so wasn't obvious.

I'm not sure how to do this. We could:
  1. Parse the output of 'cargo -V', looking at either the version number or the date.
  2. Run 'cargo --help', and search for the string "--frozen".
  3. Drop the use of "stable" and instead specify a fixed Rust version (e.g. 1.37.0). elf2tab would stick  with "stable", and presumably would receive the new version after rustup updates the toolchain.

#1 would require us to periodically update the supported version. #2 could be automated by using a wrapper binary that does that check before invoking cargo -- but the build system for that wrapper binary (or should it be a Python script? I don't want to pick up too many build dependencies) has the same issue.

Perhaps we should do #3?

phil-levis-google:
I think Tock has a way of handling this -- check out Makefile.common in boards/.

jrvanwhy:
> I think Tock has a way of handling this -- check out Makefile.common in boards/.

That's method 1. I implemented it in #129

PR #128, opened 2019-10-22
Cleanup libgolf2 and AES syscalls.

phil-levis-google:
This cleans up the required libgolf2 includes and its AES syscalls. 

Before this, applications that wanted to use certain libgolf2 syscalls (e.g., personality) needed to include their own copies of required header files (e.g., p256.h, storage.h). This was a bad dependency from the library to an application. This PR moves those properly into libgolf2.

It also fixes the system call semantics for the AES drivers, updating the aes_test app accordingly. The userspace API was incorrectly using a prior syscall API from the first h1b AES implementation, before the standard AES API Tock has now. This caused it to confused return values of bytes processed with error codes. As a result, calls to multi-block operations would incorrectly terminate early and only operate on the first block.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
bf7f413b9fc6ef86589b6e5983f133b75d3073ca
git status
On branch libgolf2-cleanups
Your branch is up to date with 'origin/libgolf2-cleanups'.

```


userspace/aes_test/Build.mk
```diff
@@ -0,0 +1,15 @@
+# Copyright 2018 Google LLC
```
jrvanwhy: The copyright headers on new files should be 2019.

jrvanwhy (approved):

userspace/aes_test/Build.mk
phil-levis-google: Fixed.

PR #132, opened 2019-10-31
Userspace support for nonvolatile counter and syscall driver fixes.

phil-levis-google:
This PR adds a userspace C library for access to the nonvolatile counter. It also fixes two bugs in the system call driver (to be able to support the library).

The first bug relates to the `wants_increment` flag. This flag indicates that a userspace process wants to increment the counter. It's used to scan across the processes on a completion event, and if one of them has the flag set, start an increment on its behalf. This flag was being set both when a process was enqueued as well as when it could immediately increment; by setting it when it could immediately increment, the driver would cause it to increment twice: once immediately, then once again when the increment callback came in. The fix is to only set it when the process is enqueued.

The second bug related to responding to a successful increment callback. Rather than check for a SuccessWithValue and using the value to set the software state of the counter, the implementation was checking for a SUCCESS. This was never the case (since it's SuccessWithValue), such that increments never seemed to succeed. The fix is to handle SuccessWithValue and set the software counter state based on the passed value.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
c2d48a3ae62bcd3066fa8c76c4ab088c18ab2782
git status
On branch nvcounter-userspace
Your branch is up to date with 'origin/nvcounter-userspace'.

```


golf2/src/main.rs
```diff
@@ -431,20 +431,22 @@ pub unsafe fn reset_handler() {
...
-    nvcounter.set_client(nvcounter_test);
+    nvcounter.set_client(nvcounter_test); */
 
     // dcrypto_test::run_dcrypto();
     //    rng_test::run_rng();
     //flash_test.run();
     //nvcounter_test.run();
 
     // Uncomment to initialize NvCounter
+    nvcounter.set_client(nvcounter_syscall);
```
jrvanwhy, line 446: This is actually done on line 297, which is more consistent with the rest of this file. Commenting line 438 is what's necessary here (which you did as well).

golf2/src/nvcounter_syscall.rs
```diff
@@ -55,7 +55,9 @@ impl<'c, C: NvCounter<'c>> NvCounterSyscall<'c, C> {
     /// value between 0 and the previous value.
     #[allow(unused)]
     pub fn initialize(&self) {
+        debug!("Initializing NvCounterSyscall.");
```
jrvanwhy, line 58: Hmm, I think the name `initialize()` is misleading here. You're really initializing the counter, not the NvCounterSyscall interface to the counter.

Perhaps this message should be "NvCounterSyscall initializing counter".

jrvanwhy (approved):

PR #146, opened 2020-03-25
Add "papa" board.

osenft:
```
Tested with TANGO_BOARD_FOR_TEST=papa

----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
93a47e57bf1e6122644d516ba8e1b981af91f2c9
git status
On branch add-papa-board
```

### DIFF
--- a/papa/Cargo.toml
+++ b/papa/Cargo.toml
@@ -1,0 +1,40 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[package]
+name = "papa"
+version = "0.1.0"
+authors = ["Oskar Senft <osk@google.com>"]
+build = "build.rs"
+edition = "2018"
+
+[profile.dev]
+panic = "abort"
+lto = false
+opt-level = "z"
+debug = true
+
+[profile.release]
+panic = "abort"
+lto = true
+opt-level = "z"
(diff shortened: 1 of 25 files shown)


papa/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+lto = true
+opt-level = "z"
+debug = true
+
+[dependencies]
+capsules = { path = "../third_party/tock/capsules" }
+kernel = { path = "../third_party/tock/kernel" }
+cortexm3 = { path = "../third_party/tock/arch/cortex-m3" }
+h1b = { path = "../h1b" }
+
```
jrvanwhy, line 39: Please remove the blank lines at the end of the file.

papa/Xargo.toml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2020 Google LLC
```
jrvanwhy, line 1: I don't think we use Xargo anymore, so Xargo.toml can be removed.

papa/src/main.rs
```diff
@@ -0,0 +1,465 @@
...
+use kernel::hil;
+use kernel::hil::entropy::Entropy32;
+use kernel::hil::rng::Rng;
+use kernel::mpu::MPU;
+
+use h1b::crypto::dcrypto::Dcrypto;
+use h1b::hil::flash::Flash;
+use h1b::nvcounter::{FlashCounter,NvCounter};
+use h1b::timels::Timels;
+use h1b::usb::{Descriptor, StringDescriptor};
```
jrvanwhy, line 58: Do you want USB to be part of Papa? It doesn't seem relevant.

papa/src/nvcounter_syscall.rs
```diff
@@ -0,0 +1,191 @@
+// Copyright 2020 Google LLC
```
jrvanwhy, line 1: `NvCounter` may not be relevant either.

jrvanwhy (commented):
I think several of these things should be in a crate shared between golf2 and papa. Specifically, AES, dcrypto, and personality seem like they are chip-specific syscall interfaces rather than board-specific interfaces.

papa/src/main.rs
osenft: Oops, good catch. I meant to remove that.

PR #156, opened 2020-05-19
Sandbox invocations of `cargo` and Tock's makefiles in the build system

jrvanwhy:
Also, fix issues discovered while testing the sandbox.

`cargo` and Tock's makefiles both like reaching out to the internet when they can't find things locally. We do not want our build system to depend on the network. Further, Tock's makefiles will automatically try to install things, which I'd prefer to avoid. This adds a bubblewrap-based sandbox which prevents them from accessing the network and from touching files outside the tock-on-titan directory.

I had previously given cargo an invalid proxy definition to prevent it from reaching out to the internet. I removed that configuration, as it is not sandboxed in a nicer manner. Now, hand invocations of cargo (invocations not performed by our makefiles) can contact the internet, but the build system's invocations must remain offline.

I also fixed the following issues, discovered while testing this sandbox:
  1. The runner was not using the correct build directory (discovered by the sandbox itself;
     cargo is unable to write in runner/target with the sandbox in place).
  2. There were redundant invocations of Tock's makefiles while building C apps. The makefiles
     are not board-specific, but they were referenced in a board-specific context, so they
     were built for each board. This caused race conditions that causes flaky failures with
     -j3 and up. I de-duplicated these invocations, making the build process faster and
     more reliable.
  3. The invocation of Tock's makefiles for building C apps wasn't correctly forwarding the
     jobserver invocation.

I confirmed this commit does not have a negative effect on build times. Here are the build times on my system for `make -j12 build check doc localtests` as of current master (measured using `time`):

real    1m20.997s
user    8m59.246s
sys     0m20.293s

and here are the build times as of this commit (note that `time` loses the ability to measure system and user time):

real    1m20.338s
user    0m1.169s
sys     0m0.316s

Open question: how should we document the new dependency? We don't currently have any "setup" docs (in part because several of the setup steps reference non-publicly-available tools).

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
db5da72a215ed7f0fc6d23813c811cc913906d95
git status
On branch sandbox-build
```


Makefile
```diff
@@ -66,4 +79,11 @@ cargo_version_check:
 		     exit 1; \
 		fi
 
+# A target that sets up directories the bwrap sandbox needs.
+.PHONY: sandbox_setup
+sandbox_setup:
+	mkdir -p build
```
osenft, line 85: Do we need to explicitly specify the parent directory here or is it guaranteed that we're in the right directory?

osenft (approved):

Makefile
jrvanwhy: `make` always executes actions from the same directory as the Makefile itself. e.g. if you do `make -C a/b/c/d/Makefile` to run a Makefile in a directory different from the present working directory, `make` will switch to `a/b/c/d` before running its actions.

So yes, this is guaranteed by `make`.

PR #158, opened 2020-07-10
Fix build to work from scratch.

osenft:
- Create missing Cargo.lock files.
- Add missing bind parameters for bwrap.
- Set libtock-c $(LIBNAME)_BUILDDIR variables.

This requires libtock-c at commit
c186cfcd520c669700a469daf776d4266137aef4 or later.

Confirmed that `make prtest` still runs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+.PHONY: all
+all: build
+
 # Subdirectories containing Build.mk files.
 BUILD_SUBDIRS := kernel runner third_party tools userspace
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -43,6 +58,10 @@ check: $(addsuffix /check,$(BUILD_SUBDIRS))
 # No need to recurse into most directories, as rm does that for us.
 .PHONY: clean
 clean: kernel/clean userspace/clean
+	rm -f runner/Cargo.lock
+	rm -f third_party/libtock-rs/Cargo.lock
+	rm -f third_party/rustc-demangle/Cargo.lock
+	rm -f tools/Cargo.lock
```
jrvanwhy: `runner/`, `third_party/`, and `tools/` already have `Build.mk` files. We should add `clean` targets to those directories that invoke `cargo clean`, then make this `clean` depend on those (in the same way it depends on `userspace/clean` and `kernel/clean`).

jrvanwhy (changes requested):

Makefile
osenft: Good point, thanks for suggesting that. Fixed. See additional commit.

jrvanwhy (approved):

PR #160, opened 2020-07-10
Add a GitHub action that runs `make check localtests`.

jrvanwhy:
Unfortunately, we cannot run `make build` without access to a Google-internal code signing tool. `make check localtests` is the next-best option available to us that doesn't require making large changes to the build system.

I added a `make setup` target that sets up the corresponding toolchains.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
948e96c7ddd89682fab5c346bd8726152bdf967b
git status
On branch action
Your branch is up to date with 'origin/action'.

```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,55 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: ci
+on: pull_request
+
+jobs:
+  build:
+    # Using ubuntu-latest can cause breakage when ubuntu-latest is updated to
+    # point at a new Ubuntu version. Instead, explicitly specify the version, so
+    # we can update when we need to. This *could* break if we don't update it
+    # until support for 18.04 is dropped, but it is likely we'll have a reason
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,55 @@
...
+        run: sudo apt-get install bubblewrap
+
+      # Runs the `check` and `localtests` make actions. We cannot build this
+      # repository in GitHub Actions as building requires non-publicly-available
+      # code signing tools, so this is the best we can do (at least for now --
+      # theoretically we could build up until the code signing step but the
+      # makefiles are not structured for that yet). I use -j2 because the VMs
+      # that GitHub Actions uses have 2 logical CPUs.
+      - name: Check and local tests
+        run: make -C "${GITHUB_WORKSPACE}" -j2 check localtests
```
osenft, line 55: How about just using `-j`?

osenft (approved):

.github/workflows/ci.yaml
jrvanwhy: I went ahead and ran a benchmark with three `-j` options:

- `-j1`: [3m 55s](https://github.com/jrvanwhy/tock-on-titan/runs/866309162)
- `-j2`: [3m 26s](https://github.com/jrvanwhy/tock-on-titan/runs/866307638)
- `-j3`: [3m 33s](https://github.com/jrvanwhy/tock-on-titan/runs/866310002)

`-j2` appears to be the most efficient value by a bit (though it is unclear what the measurement noise is -- I don't want to run these tests too many times as AFAIK GitHub is donating these resources to OSS). I wouldn't expect higher `-j` values to perform any better than `-j3`, as it'll just cause more context switching and memory use-related overhead (and I would expect an OOM from `-j`).

jrvanwhy: After sending the previous message, I realized that `-j` may not trigger `cargo`'s concurrency, and may therefore behave differently than `-jN` for `N > 3`. Therefore I ran a build with `-j`. It took [3m 50s](https://github.com/jrvanwhy/tock-on-titan/runs/866369536), which is slightly faster than `-j1` but slower than `-j2`.

PR #161, opened 2020-07-13
Add spiutils lib.

osenft:
This library will be used both by Tock kernel and in userspace.


shared-lib/spiutils/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+
```
jrvanwhy: Please remove the blank line after `[package]`.

shared-lib/spiutils/Cargo.toml
```diff
@@ -0,0 +1,10 @@
...
+name = "spiutils"
+version = "0.1.0"
+authors = [ "Oskar Senft <osk@google.com>" ]
+license = "MIT"
```
jrvanwhy: This is `Apache-2.0` licensed, not `MIT` licensed.

shared-lib/spiutils/Cargo.toml
```diff
@@ -0,0 +1,10 @@
...
+keywords = ["spi-utils"]
```
jrvanwhy: `spi-utils` isn't really a good keyword. I don't expect this to be uploaded to crates.io (do you?) so I don't think we need to specify keywords.

shared-lib/spiutils/Cargo.toml
```diff
@@ -0,0 +1,10 @@
...
+description = """
+Utilities for SPI
```
jrvanwhy, line 7: I'm not sure this description is very good -- this crate is probably going to be somewhat Titan-specific, correct?

I think we can remove it, unless you expect this to be uploaded to `crates.io`.

shared-lib/spiutils/src/cmds.rs
```diff
@@ -0,0 +1,86 @@
+/// SPI flash op codes
+#[repr(u8)]
+#[allow(dead_code)]
+pub enum OpCodes {
+    /// No operation
+    Nop = 0x00,
+
+
```
jrvanwhy, line 9: Too many blank lines here? Probably should just be 1.

shared-lib/spiutils/src/cmds.rs
```diff
@@ -0,0 +1,86 @@
+/// SPI flash op codes
+#[repr(u8)]
+#[allow(dead_code)]
```
jrvanwhy: `#[allow(dead_code)]` shouldn't be necessary here.

jrvanwhy (changes requested):

shared-lib/spiutils/Cargo.toml
osenft: I was expecting the crate to be independent from Titan. Two things I know I want in there:
- SPI flash commands (done)
- SPI flash ID generation like JEDEC ID and SFDP.

shared-lib/spiutils/Cargo.toml
osenft: Done.

shared-lib/spiutils/Cargo.toml
osenft: Changed. I wasn't sure what this should be set to, I think I copied it from another lib. Thanks for clarifying!

shared-lib/spiutils/Cargo.toml
jrvanwhy: Perhaps the crate name should be something like `spi_flash_decls`? `spiutils` isn't descriptive enough for `crates.io`.

shared-lib/spiutils/src/cmds.rs
osenft: I was using that to visually distinguish commands from each other. I added some additional comment in.

shared-lib/spiutils/src/cmds.rs
osenft: Ah yes, good catch, thank you.

shared-lib/spiutils/Cargo.toml
osenft: Removed.

shared-lib/spiutils/Cargo.toml
jrvanwhy: Do you want to give this a name more suitable for `crates.io` or submit as-is?

shared-lib/spiutils/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "spiutils"
+version = "0.1.0"
+authors = [ "Oskar Senft <osk@google.com>" ]
+license = "Apache-2.0"
+description = """
+Utilities for SPI
+"""
```
jrvanwhy, line 8: Cargo.toml is missing an EOL character at its end.

jrvanwhy (approved):
Please replace the missing EOL character in `Cargo.toml`, otherwise LGTM.

PR #165, opened 2020-07-17
Use bubblewrap to replace elf2tab's Cargo.lock with a new empty one.

jrvanwhy:
This effectively disables cargo's reproducible build mechanism, allowing us to modify elf2tab's dependencies. This will allow us to remove many dependencies that we don't need (such as winapi) by patching dependencies' Cargo.toml files. It will also improve our ability to deduplicate different versions of a crate, because we will no longer need to match crate checksums to elf2tab's Cargo.lock.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
73537c5c76010d739b3e20f81546a643f57125cb
git status
On branch cargo-lock-hack
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,8 @@ BUILD_SUBDIRS := kernel runner third_party tools userspace
 BWRAP := bwrap                                                               \
          --ro-bind / /                                                       \
          --bind "$(CURDIR)/build" "$(CURDIR)/build"                          \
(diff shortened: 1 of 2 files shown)


osenft (approved):
Can you please test that this work from a fresh clone?

jrvanwhy:
> Can you please test that this work from a fresh clone?

Done, it works for me in a fresh checkout. That said, the `make check` step of the CI should catch any issues caused by this change.

PR #170, opened 2020-07-29
Add Manticore to Tock-on-Titan

osenft:
This adds both [Manticore](https://github.com/lowRISC/manticore/) and its dependencies to `third_party`. Dependencies are added with version numbers in the directory names to allow for different versions of the same library to be present. Dependencies have already been de-duped, so that it uses existing dependencies without version number in the directory name.


jrvanwhy:
Here is the dependency tree of manticore (no dev or test dependencies included, default features off):

```
└── manticore v0.0.1 (/usr/local/google/home/jrvanwhy/tock-on-titan/third_party/manticore)
    ├── arrayvec v0.5.1
    ├── bitflags v1.2.1
    ├── byteorder v1.3.4
    ├── static_assertions v1.1.0
    └── zerocopy v0.3.0
        ├── byteorder v1.3.4 (*)
        └── zerocopy-derive v0.2.0
            ├── proc-macro2 v1.0.19 (*)
            ├── syn v1.0.36 (*)
            └── synstructure v0.12.4
                ├── proc-macro2 v1.0.19 (*)
                ├── quote v1.0.7 (*)
                ├── syn v1.0.36 (*)
                └── unicode-xid v0.2.1 (*)
```

You've imported several large crates, including `ring` and `web-sys`, which `manticore` does not depend on (at least not with default features off). As far as I can tell, an app that uses `manticore` will build fine without these dependencies.

Please remove the unnecessary dependencies. Here is an example: https://github.com/osenft/tock-on-titan/pull/1. Note we'll probably want to rebase if necessary to prevent the large dependencies from being a part of tock-on-titan's permanent git history.

osenft:
> Here is the dependency tree of manticore (no dev or test dependencies included, default features off):
> ...

Oh man, thank you! I accidentally had added the dependencies with default features turned on. I corrected the commits. PTAL.

jrvanwhy:
This PR (as-is) introduces new versions of `proc-macro2`, `quote`, `syn`, and `unicode-xid`, all of which we already vendored. `manticore` seems to work fine with the already-vendored versions of those dependencies (it doesn't need the newer versions you added).

Please remove `proc-macro2-1.0.19`, `quote-1.0.7`, `syn-1.0.36`, and `unicode-xid-0.2.1`. If you'd like, I opened https://github.com/osenft/tock-on-titan/pull/3 with the changes I tested which you can merge in.

.gitmodules
```diff
@@ -18,3 +18,6 @@
 [submodule "third_party/tock"]
 	path = third_party/tock
 	url = https://github.com/tock/tock.git
+[submodule "third_party/manticore"]
```
jrvanwhy, line 21: Nitpick: Please add a blank line before the manticore submodule definition to be consistent with the rest of this file.

jrvanwhy (approved):

PR #175, opened 2020-09-11
Add ux library as dependency for shared-lib/spiutils.

osenft:


third_party/ux-0.1.3/src/lib.rs
```diff
@@ -0,0 +1,997 @@
...
+
+
+mod lib {
+    pub mod core {
+        #[cfg(feature="std")]
+        pub use std::*;
+        #[cfg(not(feature="std"))]
+        pub use core::*;
+    }
+}
```
jrvanwhy, line 22: I know you didn't write this code, but my opinion of lines 15 - 22 is "yuck".

Idk why Rust developers are so insistent on using the types in `std` whenever possible. If a crate optionally supports `no_std`, then it should just refer to things in `core` by their `core` path, not do this renaming.

jrvanwhy (approved):

PR #183, opened 2020-09-16
Add support for program and erase

osenft:
```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
84fd05aa7188fd34eeb922d0dba38ab226683350
git status
On branch add-write-erase-support
Your branch is up to date with 'origin/add-write-erase-support'.

```

### DIFF
--- a/userspace/otpilot/src/main.rs
+++ b/userspace/otpilot/src/main.rs
@@ -29,12 +29,13 @@ use libtock::result::TockResult;
 
 use manticore::crypto::rsa;
 use manticore::hardware;
-use manticore::io::Cursor;
+use manticore::io::Cursor as ManticoreCursor;
 use manticore::protocol::capabilities::*;
 use manticore::protocol::device_id;
 use manticore::server::pa_rot::{PaRot, Options};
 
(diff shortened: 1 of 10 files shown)


userspace/otpilot/src/main.rs
```diff
@@ -225,29 +226,113 @@ impl<'a> SpiProcessor<'a> {
...
+                }
+            }
+
+            if let Err(why) = tx_cursor.write_bytes(&data) {
+                writeln!(console, "Host: Could not store data: {:?}", why);
+                return Err(FromWireError::OutOfRange);
             }
+
+            tx_len = tx_cursor.consumed_len()
         }
```
jrvanwhy, line 258: Why are lines 243-263 in an indented block?

osenft: We need to lose `tx_cursor` since it holds a `&mut` on `tx_buf`. We need to get a new `&mut` on `tx_buf` in line 268.

osenft: I moved the code into its own function, I think this is cleaner now.

jrvanwhy (approved):

PR #184, opened 2020-09-16
Add error handling to OTPilot

osenft:
This PR builds on top of https://github.com/google/tock-on-titan/pull/183.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
5199c0083d9eb1df1820a40a308279dee415f74d
git status
On branch otpilot-error-handling
Your branch is up to date with 'origin/otpilot-error-handling'.

nothing to commit, working tree clean
```

PR #185, opened 2020-09-17
Add support for JEDEC ID and SFDP

osenft:
This PR builds on top of https://github.com/google/tock-on-titan/pull/184.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
b75f21934e4855c88f32fae3ab49559ea93a00f6
git status
On branch jedec
```

### DIFF
--- a/kernel/h1/src/spi_device.rs
+++ b/kernel/h1/src/spi_device.rs
@@ -844,7 +844,7 @@ impl SpiDevice for SpiDeviceHardware {
             AddressMode::ThreeByte => self.registers.eeprom_ctrl.modify(EEPROM_CTRL::ADDR_MODE::CLEAR),
             AddressMode::FourByte => self.registers.eeprom_ctrl.modify(EEPROM_CTRL::ADDR_MODE::SET),
         }
-        debug!("set_address_mode: {:?}", address_mode);
+        //debug!("set_address_mode: {:?}", address_mode);
     }
 
     fn get_address_mode(&self) -> AddressMode {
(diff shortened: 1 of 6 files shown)


osenft:
I just rebased this PR to #184. PTAL.

kernel/h1/src/spi_device.rs
```diff
@@ -16,6 +16,10 @@ use kernel::ReturnCode;
 use spiutils::protocol::flash::AddressMode;
 use spiutils::protocol::flash::OpCode;
 
+fn get_data_byte_or_default(data: &[u8], idx: usize, default: u8) -> u8 {
+    if idx < data.len() { data[idx] } else { default }
```
jrvanwhy: This can be `data.get(idx).unwrap_or(default)`

userspace/otpilot/src/main.rs
```diff
@@ -160,73 +163,113 @@ impl rsa::Builder for NoRsa {
...
+    Format(core::fmt::Error),
+}
+
+impl From<FromWireError> for SpiProcessorError {
+    fn from(err: FromWireError) -> Self {
+        SpiProcessorError::FromWire(err)
+    }
+}
+
+impl From<core::num::TryFromIntError> for SpiProcessorError {
```
jrvanwhy: I'm not sure I like the use of `From<core::num::TryFromIntError>` to create a `SpiProcessorError::FromWire`. It doesn't feel like a clean abstraction. As far as I can tell, it is only used in one location.

Perhaps we should remove this `From` implementation and replace `u16::try_from(tx_cursor.consumed_len())?` with `u16::try_from(tx_cursor.consumed_len()).map_err(|_| FromWireError::OutOfRange.into())?`

userspace/otpilot/src/sfdp.rs
```diff
@@ -0,0 +1,566 @@
...
+
+    for idx in 0..sfdp.len() {
+        data[idx] = sfdp[idx];
+    }
+    for idx in sfdp.len()..data.len() {
+        data[idx] = !0;
+    }
+
+    Ok(())
+}
```
jrvanwhy: Nitpick: the last line of this file is missing an EOL character, please add one.

jrvanwhy (changes requested):

kernel/h1/src/spi_device.rs
osenft: Ha, yeah, I did not know that, thank you! I tried to actually merge the get_data_byte_or_default back into the code, but the resulting line was very hard to read due to references and dereferencing and the resulting need of parantheses. So I just updated the helper function and added a comment.

userspace/otpilot/src/main.rs
osenft: Good point. The other cases should be clean. Thanks for pointing this out!

userspace/otpilot/src/sfdp.rs
osenft: One day I'll figure out how to make Sublime do that automatically ... sigh.

jrvanwhy (approved):

PR #189, opened 2020-09-29
Split SPI writes incoming from SPI device before sending to SPI host

osenft:
```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
fc4966c9b69916ebdb0b64075a97f11094bdd3aa
git status
On branch split-spi-writes
Your branch is up to date with 'origin/split-spi-writes'.

```

### DIFF
--- a/shared-lib/spiutils/src/protocol/flash.rs
+++ b/shared-lib/spiutils/src/protocol/flash.rs
@@ -209,22 +209,32 @@ impl<'a> OpCode {
 
 const DUMMY_BYTE_VALUE: u8 = 0xff;
 
+/// Error used when address cannot be converted.
+pub struct AddressConversionError;
+
 /// Address in SPI flash protocol header
 pub trait Address: BeInt + Into::<u32> {
     /// Convert from a u32.
     /// Note: This will become redundant once types from `ux` crate implement
     /// `core::convert::TryFrom`.
-    fn try_from(val: u32) -> Result<Self, core::num::TryFromIntError>;
+    fn try_from(val: u32) -> Result<Self, AddressConversionError>;
 }
 
 impl Address for ux::u24 {
-    fn try_from(val: u32) -> Result<Self, core::num::TryFromIntError> {
-        Ok(ux::u24::new(val))
+    fn try_from(val: u32) -> Result<Self, AddressConversionError> {
+        // ux::[type]::new just assert!s that the val fits, which results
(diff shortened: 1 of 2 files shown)


shared-lib/spiutils/src/protocol/flash.rs
```diff
@@ -208,6 +209,26 @@ impl<'a> OpCode {
...
+pub trait Address: BeInt + Into::<u32> {
+    /// Convert from a u32.
+    /// Note: This will become redundant once types from `ux` crate implement
+    /// `core::convert::TryFrom`.
+    fn try_from(val: u32) -> Result<Self, core::num::TryFromIntError>;
+}
+
+impl Address for ux::u24 {
+    fn try_from(val: u32) -> Result<Self, core::num::TryFromIntError> {
+        Ok(ux::u24::new(val))
```
jrvanwhy: I would expect this to return an error if `val > u24::max_value()`. Does this need clarification or fixing?

shared-lib/spiutils/src/protocol/flash.rs
```diff
@@ -221,15 +242,23 @@ pub struct Header<AddrType> {
...
-        let mut address: Option<AddrType> = None;
-        if opcode.has_address() {
-            address = Some(r.read_be::<AddrType>()?);
-        }
+        let address =
+            if opcode.has_address() {
+                Some(r.read_be::<AddrType>()?)
+            } else {
+                None
+            };
```
jrvanwhy: This pattern can be written in 4 lines rather than 6 with `match`:

```
let address = match opcode.has_address() {
    true => Some(r.read_be::<AddrType>()?),
    false => None,
};
```

jrvanwhy (changes requested):

shared-lib/spiutils/src/protocol/flash.rs
osenft: Oh, that was a good call! On inspection, I found that ux::[type]::new just asserts that the value fits and panics if it doesn't. I replaced the TryFromIntError with a custom error since TryFromIntError cannot be instantiated from outside the crate that it's defined in.

shared-lib/spiutils/src/protocol/flash.rs
osenft: One day ... one day I'll learn Rust!

jrvanwhy (approved):

PR #190, opened 2020-11-18
SPI device on Papa board starts up in 3-byte address mode.

tjshep:
"make build localtests"  did run to completion with exit status zero.


tjshep:
@osenft

osenft (approved):

PR #193, opened 2021-02-10
Add some detail to Readme.md; Don't fail build when one of TANGO_CODESIGNER{,_KEY} is not set.

dann21:

Well, not sure that the Makefile change is a good idea -- see commit comment.

`make prtest` doesn't pass, regardless of whether `TANGO_CODESIGNER{,_KEY}` are set, but it fails differently (and earlier, I think) when they're not set.  Which reinforces my nervousness about the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,3 +4,38 @@ This repository contains ports of Tock OS (https://www.tockos.org) to Titan
 chips.
 
 This is not an officially supported Google product.
+
+
+## Getting started
+
+### Clone the repo
+
+Get the source:
+
+```shell
+git clone --recursive https://github.com/google/tock-on-titan.git
+```
+
+### Get the tools and libs to build the code
+
+Download Rust
+
+```shell
+cd tock-on-titan
+curl https://sh.rustup.rs -sSf | sh
+```
+
+Configure Rust
+
+```shell
+make setup
+```
+
+### Build all boards and apps
+
+```shell
+make
+```
+
+Note that if one of TANGO_CODESIGNER{,_KEY} is not set, then signed artifacts
+will not be created.
--- a/userspace/Build.mk
+++ b/userspace/Build.mk
@@ -130,11 +130,16 @@ build/userspace/$(APP)/$(BOARD)/full_image: \
 	arm-none-eabi-objcopy --update-section \
 		.apps=build/userspace/$(APP)/cortex-m3/cortex-m3.tbf \
 		build/userspace/$(APP)/$(BOARD)/unsigned_image
-	$(TANGO_CODESIGNER) --b --input build/userspace/$(APP)/$(BOARD)/unsigned_image \
-		--key=$(TANGO_CODESIGNER_KEY) \
-		--output=build/userspace/$(APP)/$(BOARD)/signed_image
-	cat $(TANGO_BOOTLOADER) build/userspace/$(APP)/$(BOARD)/signed_image \
-		> build/userspace/$(APP)/$(BOARD)/full_image
+	if [ -n "${TANGO_CODESIGNER}" -a -n "${TANGO_CODESIGNER_KEY}" ]; then \
+		$(TANGO_CODESIGNER) --b --input build/userspace/$(APP)/$(BOARD)/unsigned_image \
+			--key=$(TANGO_CODESIGNER_KEY) \
+			--output=build/userspace/$(APP)/$(BOARD)/signed_image; \
+		cat $(TANGO_BOOTLOADER) build/userspace/$(APP)/$(BOARD)/signed_image \
+			> build/userspace/$(APP)/$(BOARD)/full_image; \
+	else \
+		echo "***** Can't create signed_image -- one of CODESIGNER{,_KEY} is empty "; \
+		echo "*****   -- hope that's OK."; \
+	fi
 
 endef
 
@@ -283,11 +288,16 @@ build/userspace/$(APP)/$(BOARD)/full_image: \
 	arm-none-eabi-objcopy --update-section \
 		.apps=build/userspace/$(APP)/$(BOARD)/app.tbf \
 		build/userspace/$(APP)/$(BOARD)/unsigned_image
-	$(TANGO_CODESIGNER) --b --input build/userspace/$(APP)/$(BOARD)/unsigned_image \
-		--key=$(TANGO_CODESIGNER_KEY) \
-		--output=build/userspace/$(APP)/$(BOARD)/signed_image
-	cat $(TANGO_BOOTLOADER) build/userspace/$(APP)/$(BOARD)/signed_image \
-		> build/userspace/$(APP)/$(BOARD)/full_image
+	if [ -n "${TANGO_CODESIGNER}" -a -n "${TANGO_CODESIGNER_KEY}" ]; then \
+		$(TANGO_CODESIGNER) --b --input build/userspace/$(APP)/$(BOARD)/unsigned_image \
+			--key=$(TANGO_CODESIGNER_KEY) \
+			--output=build/userspace/$(APP)/$(BOARD)/signed_image; \
+		cat $(TANGO_BOOTLOADER) build/userspace/$(APP)/$(BOARD)/signed_image \
+			> build/userspace/$(APP)/$(BOARD)/full_image; \
+	else \
+		echo "***** Can't create signed_image -- one of CODESIGNER{,_KEY} is empty "; \
+		echo "*****   -- hope that's OK."; \
+	fi
 
 endef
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -4,3 +4,36 @@ This repository contains ports of Tock OS (https://www.tockos.org) to Titan
...
+```shell
+cd tock-on-titan
+curl https://sh.rustup.rs -sSf | sh
+```
+
+```shell
+rustup default $(cat third_party/libtock-rs/rust-toolchain)
+rustup target add thumbv7m-none-eabi
+rustup default $(cat kernel/rust-toolchain)
+rustup target add thumbv7m-none-eabi
```
jrvanwhy: It may be better to advise running `make setup` here. It is more likely to be up to date than a README. It also doesn't change the user's default rust toolchain.

jrvanwhy:
Does it make sense to have two primary build targets -- one that builds up until the signing step and stops, and one that goes all the way?

It would be useful for CI, as our only sign is GitHub Actions which doesn't have a signing key.

README.md
dann21: Ooh -- cool.  Will fix; will also fix osk@'s internal howto doc.

dann21: Done.

dann21:
> Does it make sense to have two primary build targets -- one that builds up until the signing step and stops, and one that goes all the way?
> 
> It would be useful for CI, as our only sign is GitHub Actions which doesn't have a signing key.

Not sure what's right here.  `make prtest` doesn't work in my environment (running on my laptop), when `TANGO_CODESIGNER{,_KEY}` are unset.  When they are set to use a test file, it fails in a different way.  No idea if it works when set for the real boards (all of which require Gnubby), because I don't know how to make them work.

Until I get to the point where I can test this kind of change, I probably shouldn't be making it.  :-)

But the README change should be fine, as far as it goes.  Except that, without _some_ sort of Makefile change, nobody who follows the README is going to be happy.

jrvanwhy:
Yeah, `make prtest` runs several tests on a physical board, which requires working codesigning. There's not really a good way around that. I spent a lot of time trying to find ways to spin up some sort of remote test invocation infrastructure, but it is a ton of work to do so.

userspace/Build.mk
```diff
@@ -130,9 +130,16 @@ build/userspace/$(APP)/$(BOARD)/full_image: \
...
+		$(TANGO_CODESIGNER) --b --input build/userspace/$(APP)/$(BOARD)/unsigned_image \
+			--key=$(TANGO_CODESIGNER_KEY) \
+			--output=build/userspace/$(APP)/$(BOARD)/signed_image; \
+	else \
+		echo "***** Substituting unsigned_image for signed_image -- one of"; \
+		echo "*****   CODESIGNER{,_KEY} is empty-- hope that's OK."; \
+		cp -a build/userspace/$(APP)/$(BOARD)/unsigned_image \
+			build/userspace/$(APP)/$(BOARD)/signed_image; \
+	fi
 	cat $(TANGO_BOOTLOADER) build/userspace/$(APP)/$(BOARD)/signed_image \
```
osenft: I think you should pull the `cat ... > .../full_image` into the `if`-clause, as well. Pretending to generate a signed_image and with that the full_image is confusing.

userspace/Build.mk
```diff
@@ -130,9 +130,16 @@ build/userspace/$(APP)/$(BOARD)/full_image: \
...
-	$(TANGO_CODESIGNER) --b --input build/userspace/$(APP)/$(BOARD)/unsigned_image \
-		--key=$(TANGO_CODESIGNER_KEY) \
-		--output=build/userspace/$(APP)/$(BOARD)/signed_image
+	if [ -n "${TANGO_CODESIGNER}" -a -n "${TANGO_CODESIGNER_KEY}" ]; then \
+		$(TANGO_CODESIGNER) --b --input build/userspace/$(APP)/$(BOARD)/unsigned_image \
+			--key=$(TANGO_CODESIGNER_KEY) \
+			--output=build/userspace/$(APP)/$(BOARD)/signed_image; \
+	else \
+		echo "***** Substituting unsigned_image for signed_image -- one of"; \
+		echo "*****   CODESIGNER{,_KEY} is empty-- hope that's OK."; \
```
osenft: Add space after `empty` and before `--`.

userspace/Build.mk
```diff
@@ -283,9 +290,16 @@ build/userspace/$(APP)/$(BOARD)/full_image: \
 	arm-none-eabi-objcopy --update-section \
 		.apps=build/userspace/$(APP)/$(BOARD)/app.tbf \
 		build/userspace/$(APP)/$(BOARD)/unsigned_image
-	$(TANGO_CODESIGNER) --b --input build/userspace/$(APP)/$(BOARD)/unsigned_image \
-		--key=$(TANGO_CODESIGNER_KEY) \
-		--output=build/userspace/$(APP)/$(BOARD)/signed_image
+	if [ -n "${TANGO_CODESIGNER}" -a -n "${TANGO_CODESIGNER_KEY}" ]; then \
```
osenft, line 291: The copy&paste is obviously not desirable, but you're not introducing it, so I think it's ok.

README.md
```diff
@@ -4,3 +4,35 @@ This repository contains ports of Tock OS (https://www.tockos.org) to Titan
...
+
+
+## Getting started
+
+### Clone the repo
+
+Get the source:
+
+```shell
+git clone --recursive git@github.com:google/tock-on-titan.git
```
osenft: I'd use the HTTP URL here (`https://github.com/google/tock-on-titan.git`) instead of SSH.

README.md
```diff
@@ -4,3 +4,35 @@ This repository contains ports of Tock OS (https://www.tockos.org) to Titan
...
+
+```shell
+cd tock-on-titan
+curl https://sh.rustup.rs -sSf | sh
+```
+
+Configure Rust
+
+```shell
+make setup
```
osenft, line 31: Oh, I didn't realize we had that, cool! Does that make the manual "rustup" calls obsolete? Can we test that by purging some rust directory?

README.md
```diff
@@ -4,3 +4,35 @@ This repository contains ports of Tock OS (https://www.tockos.org) to Titan
...
+git clone --recursive git@github.com:google/tock-on-titan.git
+```
+
+### Get the tools and libs to build the code
+
+Download Rust
+
+```shell
+cd tock-on-titan
+curl https://sh.rustup.rs -sSf | sh
```
osenft, line 25: Could we also add that into the `setup` make target?

userspace/Build.mk
dann21: Hmm.  OK.  For some reason, I had convinced myself that `make build` would fail when `full_image` isn't created, but I guess I never tested it.  There _are_ targets that depend on `full_image`'s existence, but I guess `make build` doesn't need them.  So later steps will fail if we don't get a `full_image`, but I guess that's OK.

userspace/Build.mk
dann21: If I'm not creating `full_image` when one of CODESIGNER{,_KEY} is not set, there's no point in creating `signed_image` either, so I have changed the comment.

userspace/Build.mk
dann21: Yeah -- three copies of almost identical code here.  Pretty clear how to abstract them into a single `define` -- maybe I'll do that after.

README.md
dann21: Don't think so.  `make setup` does exactly what the explicit `rustup` commands did.

README.md
dann21: I'd be mildly against doing that (but only mildly).

The `curl` command actually installs a fully working (recent) version of rust.  I think it's important for the user to be aware that they have done that.

The `rustup` commands that `make setup` runs installs two different versions of the toolchain, along with installing two different versions of what looks like a `thumbv7em-none-eabi` compiler plug-in.  That feels like a different thing to me, and running `make setup` does get across the idea that something like that is going on.

That being said, if you still think that the convenience of not typing a second command outweighs the loss of transparency, I'll make the change.  Let me know.

userspace/Build.mk
dann21: Done.

userspace/Build.mk
dann21: Done.

README.md
dann21: Done.

userspace/Build.mk
dann21: I've worked up a change that abstracts those three copies into a single macro.  I will submit it after as a separate PR.

README.md
dann21: Hmm.  I think I've convinced myself that you're right about this. I will make this change in the next round.

userspace/Build.mk
dann21: See branch `dann/factor-FULL_IMAGE_TARGET` of https://github.com/dann21/tock-on-titan.git.  I will cut a PR for what's on this branch once this PR has been disposed of.

README.md
dann21: See branch `dann/add-initial-rust-download-to-setup-target of https://github.com/dann21/tock-on-titan.git.  I will cut a PR for what's on this branch once this PR has been disposed of.

osenft: Ah, good point re `rustup` installing a whole Rust environment. I guess being explicit (as you have it now) is better then. LG!

osenft (approved):
LGTM

README.md
dann21: I think we just swapped positions.  We can continue the discussion if/when I submit a PR for branch `dann/add-initial-rust-download-to-setup-target` of https://github.com/dann21/tock-on-titan.git. 

Heh.

jrvanwhy: Is the `rustup` installation a no-op if you already have `rustup` installed? I believe `make setup` can safely be run multiple times (i.e. I think it should be idempotent).

jrvanwhy (approved):

README.md
dann21: Yes.  Actually, to be fair, I'm not certain that it's a bit-for-bit no-op -- it may rerun the installation, copying new bits over the old bits.  So running it may refresh what's there.  But it won't delete anything, so it _appears_ idempotent to me.

dann21: In any case, for this purpose, I'm pretty sure it's OK.

dann21: If needed, we can carry this discussion further when I make a PR out of branch `dann/add-initial-rust-download-to-setup-target of https://github.com/dann21/tock-on-titan.git.  Coming soon.

dann21: FYI, I did try this in a separate PR.  Mt attempt failed.  See #195 for details.

PR #195, opened 2021-02-11
Add initial rust download to setup target.

dann21:
Also:
* Adjust README.md accordingly.

Signed-off-by: Dan Nussbaum <dansn@google.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,9 +86,12 @@ prtest: build devicetests localtests
 	git status
 	@echo '```'
 
-# Installs the necessary Rust toolchains
+# Download Rust (see https://www.rust-lang.org/learn/get-started --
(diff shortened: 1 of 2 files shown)


jrvanwhy (approved):

jrvanwhy (changes requested):
Nevermind on LGTM, the CI error may be a deal-breaker.

dann21:
This failed because the automated build is a non-interactive environment, and the command that I folded into `make setup` is interactive.  We could take steps to make it non-interactive, but that would lose useful flexibility, and the complexity that doing so only for the automated build would not be warranted.  Furthermore, such a change would buy us nothing -- the whole point of the change is to get rid of the gesture associated with the `curl` command -- leaving it interactive would mean that the gesture is still there.

PR #196, opened 2021-02-12
Revert "userspace/Build.mk: factor FULL_IMAGE_TARGET."

dann21:
This reverts commit c8fcd54a7c38b5998061ed5a791476e228d41c5d.

It looks like the make variable TBF_FILE is not set in a place that I
think it must be.

I thought I tested pretty thoroughly yesterday -- no idea how this
problem (which seems like not a corner case at all) made it through my
testing, or through the automated tests that are run.

But we need to have mainline working, so I need to revert before doing
anything further.

Signed-off-by: Dan Nussbaum <dansn@google.com>


dann21:
osk@ found a simple fix, so no need for the revert/reapply churn.

PR #199, opened 2021-02-25
otpilot: add banner generation

tjshep:
"make build localtests" did run to completion with exit status zero.

### DIFF
--- a/userspace/Build.mk
+++ b/userspace/Build.mk
@@ -62,6 +62,12 @@ userspace/doc: sandbox_setup
 .PHONY: userspace/localtests
 userspace/localtests: $(addsuffix /localtests,$(BUILD_SUBDIRS))
 
+# .PHONY so it happens every time, even though we do write a file at the target
+.PHONY: build/gitlongtag
+build/gitlongtag:
+	mkdir -p build
+	git describe --always --dirty --long >$@
+
 include $(addsuffix /Build.mk,$(BUILD_SUBDIRS))
 
 # ------------------------------------------------------------------------------
@@ -258,7 +264,7 @@ userspace/$(APP)/$(BOARD)/build-signed: \
 		build/userspace/$(APP)/$(BOARD)/full_image
 
 .PHONY: userspace/$(APP)/$(BOARD)/check
-userspace/$(APP)/$(BOARD)/check: sandbox_setup
+userspace/$(APP)/$(BOARD)/check: sandbox_setup build/gitlongtag
 	cd userspace/$(APP) && TOCK_KERNEL_VERSION=$(APP) $(BWRAP) cargo check \
 		--offline --release
 
@@ -266,7 +272,7 @@ userspace/$(APP)/$(BOARD)/check: sandbox_setup
 userspace/$(APP)/$(BOARD)/devicetests:
 
 .PHONY: userspace/$(APP)/$(BOARD)/doc
-userspace/$(APP)/$(BOARD)/doc: sandbox_setup
+userspace/$(APP)/$(BOARD)/doc: sandbox_setup build/gitlongtag
 	cd userspace/$(APP) && TOCK_KERNEL_VERSION=$(APP) $(BWRAP) cargo doc \
 		--offline --release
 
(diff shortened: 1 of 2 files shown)


jrvanwhy (approved):

jrvanwhy (changes requested):
Unfortunately, this PR is actually broken. The `git describe` will always show it as dirty, because `cargo` runs inside a sandbox that hides `elf2tab`'s `Cargo.lock`.

The best solution to this is to have a fix for https://github.com/rust-lang/cargo/issues/8504 to eliminate need for the sandbox hackery, but in the meantime we'll have to move the `git describe` call into `Build.mk`.

tjshep:
> Unfortunately, this PR is actually broken. The `git describe` will always show it as dirty, because `cargo` runs inside a sandbox that hides `elf2tab`'s `Cargo.lock`.
> 
> The best solution to this is to have a fix for [rust-lang/cargo#8504](https://github.com/rust-lang/cargo/issues/8504) to eliminate need for the sandbox hackery, but in the meantime we'll have to move the `git describe` call into `Build.mk`.

Thanks for your help last week.  I've just updated this PR with a workaround along the lines you suggested, but I still use a build.rs to assemble the banner.   The 'git describe' call is now in Build.mk and leaves the result in a .gitlongtag file in the top directory which is read in by build.rs.   I also preserved the call to 'git describe' in build.rs in a comment in the hopes that someday we won't need this workaround.

.gitignore
```diff
@@ -14,3 +14,4 @@
 
 Cargo.lock
 build/
+.gitlongtag
```
jrvanwhy: I feel like `.gitlongtag` should be in the `build/` directory, as it is a build artifact. Then it wouldn't need its own `.gitignore` rule.

userspace/otpilot/Build.mk
```diff
@@ -15,3 +15,16 @@
...
+
+# This is to work around a problem with calling git describe directly
+# from build.rs where it was always reporting dirty due to cargo
+# always rewriting the Cargo.lock of elf2tab .  We call git describe
+# here and leave the output in a file in the top directory, and then
+# build.rs will read this file.
+
+.PHONY: makegitlongtag
+
+makegitlongtag:
```
jrvanwhy: Even though this is a `.PHONY` target, it still produces a file, so we should probably make the action's name match the file it generates.

jrvanwhy (changes requested):
I think the use of `build.rs` is completely overkill. If your purpose is to document your previous approach for future use, then I think you should do that by putting it in the `doc/` folder or by referring to this commit (which would require that this PR is not rebased past this commit).

I believe something like this should work for defining the banner:

```rust
const BANNER: &'static str = concat!(
    env!("CARGO_PKG_NAME"), ' ',
    env!("CARGO_PKG_VERSION"), ' ',
    include_str!("../../build/gitlongtag")
);
```

.gitignore
tjshep: Yep.  In new push to this PR.

userspace/otpilot/Build.mk
tjshep: Yep.   New push to this PR does that.  (Still makes it .PHONY to make sure it happens every time.)

tjshep:
> I think the use of `build.rs` is completely overkill. If your purpose is to document your previous approach for future use, then I think you should do that by putting it in the `doc/` folder or by referring to this commit (which would require that this PR is not rebased past this commit).
> 
> I believe something like this should work for defining the banner:
> 
> ```rust
> const BANNER: &'static str = concat!(
>     env!("CARGO_PKG_NAME"), ' ',
>     env!("CARGO_PKG_VERSION"), ' ',
>     include_str!("../../build/gitlongtag")
> );
> ```

Yes, that was much easier and cleaner.  The build.rs is gone now.   Needed one more level of ../ to make what you suggested work right.    See latest push to this PR.

userspace/otpilot/Build.mk
```diff
@@ -15,3 +15,12 @@
...
 RUST_APPS_papa += otpilot
+
+# .PHONY so that it happens every time
+.PHONY: build/gitlongtag
+
+build/gitlongtag:
+	mkdir -p build
+	git describe --always --dirty --long >build/gitlongtag
+
+sandbox_setup: build/gitlongtag
```
jrvanwhy: I think it would be better to make `build/gitlongtag` a dependency of the targets that call `cargo check`, this isn't related to sandbox setup. I.e., starting at [userspace/Build.mk line 261](https://github.com/google/tock-on-titan/blob/2e3f84edf8d1f97f6da4a6bd0a3eedebde46db81/userspace/Build.mk#L261), add it as a dependency to `check`, `doc`, and `app`.

jrvanwhy (approved):

userspace/otpilot/Build.mk
tjshep: Yes, new push puts it all in `userspace/Build.mk` with the dependencies you suggested.   Seems to work correctly with those dependencies.

PR #200, opened 2021-03-10
third_party/Build.mk: sandbox elf2tab build by copying source

tjshep:
The bubblewrap workaround for elf2tab's checked-in Cargo.lock file
messing up offline builds does not work in a docker container.  So a
more straightforward workaround is to simply copy the source tree for
elf2tab out of third_party/ and into build/ where we can simply remove
the Cargo.lock file before calling cargo build.  With this change, it
should be possible to build in a suitable docker container (that
contains the necessary tool chains) by invoking 'make BWRAP='.

"make build localtests" did run to completion with exit status zero.

### DIFF
--- a/third_party/Build.mk
+++ b/third_party/Build.mk
@@ -30,8 +30,8 @@ third_party/build: build/cargo-host/release/elf2tab sandbox_setup
 third_party/build-signed: third_party/build
 
 .PHONY: third_party/check
-third_party/check: cargo_version_check sandbox_setup
-	cd third_party/elf2tab && \
+third_party/check: cargo_version_check sandbox_setup build/elf2tab
+	cd build/elf2tab && \
 		CARGO_TARGET_DIR="../../build/cargo-host" $(BWRAP) cargo check --release
 	cd third_party/libtock-rs && \
 		CARGO_TARGET_DIR="../../build/userspace/cargo" \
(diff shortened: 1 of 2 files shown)


third_party/Build.mk
```diff
@@ -70,5 +70,7 @@ third_party/localtests: cargo_version_check sandbox_setup
 
 .PHONY: build/cargo-host/release/elf2tab
 build/cargo-host/release/elf2tab: cargo_version_check sandbox_setup
-	cd third_party/elf2tab && \
+	cp -rp -t build third_party/elf2tab && \
```
jrvanwhy: Is `-t` needed here? `cp -pr third_party/elf2tab build` seems equivalent.

third_party/Build.mk
```diff
@@ -70,5 +70,7 @@ third_party/localtests: cargo_version_check sandbox_setup
...
+	rm -f build/elf2tab/Cargo.lock && \
```
jrvanwhy: Do we need to remove this? Shouldn't we remove it from `sandbox_setup` and `BWRAP` instead?

third_party/Build.mk
tjshep: Oh, just a different way, not needed.   Yes equivalent.    If we could go back in time, the destination directory target for cp should have always required the -t <destdir>  syntax.   I've been trying to teach myself to always do it the better way for many years but the old way is pretty well wired into my fingers.   So in interactive shells I still do it the old way, but when writing a shell script I try to remember to do it the new way.    Do you dislike the new way?

third_party/Build.mk
tjshep: in your sandbox_setup you made an empty file, which I would guess you probably needed because bwrap needs an actual file to map into the sandbox.   But simply removing the Cargo.lock file in third_party/elf2tab made the build work (except that the local git repository was now dirty).    If we remove it from sandbox_setup I expect bwrap will no longer solve the problem.    Oh, with this change I should clean up at least that part of sandbox_setup and remove the mapping from the bwrap command line?   Is that what you're suggesting?

third_party/Build.mk
jrvanwhy: Oh, I didn't realize that's a stylistic preference. It's fine with me as-is.

third_party/Build.mk
jrvanwhy: Yes, I am suggesting you remove `build/Cargo.lock/elf2tab` from `sandbox_setup` as well as the `bwrap` bind mount.

EDIT: I misread the code. We definitely need the `rm -f build/elf2tab/Cargo.lock` invocation. It shouldn't need `-f` though.

third_party/Build.mk
jrvanwhy: I'm not sure this will work correctly if we delete a file under `third_party/elf2tab` and then rebuild without cleaning first. Do we need a `rm -rf build/elf2tab` before the `cp` invocation?

tjshep: Yes.  Not likely to happen in a way that would break things, but indeed it could happen, so I've now added an rm -rf of build/ef2tab before it does a fresh copy.

third_party/Build.mk
tjshep: I cleaned up the build/Cargo.lock/elf2tab from the main Makefile now.    Leaving the BWRAP definition in there for now, still useful to make sure builds we do by hand don't depend on the network, and the use of BWRAP keeps builds we do by hand from sprinkling new Cargo.lock files in our source tree.

jrvanwhy (approved):

PR #204, opened 2021-03-19
papa: Add BMC_SRST_N, BMC_CPU_RST_N, SYS_RSTMON_N and BMC_RSTMON_N GPIO pins

osenft:


kernel/papa/src/main.rs
```diff
@@ -25,6 +25,9 @@ extern crate h1;
 extern crate kernel;
 extern crate cortexm3;
 
+use crate::kernel::hil::gpio::Configure;
+use crate::kernel::hil::gpio::Output;
```
jrvanwhy: These `use crate::kernel::...` statements don't look right to me. The `crate` keyword is used to import items from the current crate, not from other crates. I think these belong in the `use kernel::...` section below.

Also, they can be a single line:

```rust
use kernel::hil::gpio::{Configure, Output};
```

osenft: Ooops, yeah! How did that happen? Good catch! Fixed.

jrvanwhy (approved):

PR #206, opened 2021-03-23
Fix the app size check: `$` needs to be escaped twice, not once.

jrvanwhy:
This was probably broken when we moved the app build rules into defines. I also removed `<` from the `wc` commands as it is unnecessary.

Thank you to @dann21 for bringing this up!

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
0f0c985492214d8486e3e2309f0e0d944548df48
git status
On branch size-check-fix
Your branch is up to date with 'origin/size-check-fix'.

```


dann21 (approved):
LGTM

PR #209, opened 2021-04-05
kernel: Add support for Fuse controller.

osenft:


kernel/h1/src/fuse.rs
```diff
@@ -0,0 +1,62 @@
...
+    const fn new(base_addr: StaticRef<Registers>) -> FuseController {
+        FuseController {
+            registers: base_addr,
+        }
+    }
+}
+
+impl Fuse for FuseController {
+    fn get_dev_id(&self) -> u64 {
+        ((self.registers.dev_id0.get() as u64) << (mem::size_of::<u32>() * 8))
```
jrvanwhy: I think it would be clearer if you replaced `(mem::size_of::<u32>() * 8)` with `32`.

kernel/h1_syscalls/src/fuse.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn get_dev_id(&self, caller_id: AppId) -> ReturnCode {
+        self.apps.enter(caller_id, |app_data, _| {
+            if let Some(ref mut dev_id_buffer) = app_data.dev_id_buffer {
+                let dev_id = self.fuse.get_dev_id();
+                let val_size = mem::size_of::<u64>();
+                if dev_id_buffer.len() < val_size {
+                    return ReturnCode::ENOMEM;
+                }
+
+                byteorder::BE::write_u64(&mut dev_id_buffer.as_mut(), dev_id);
```
jrvanwhy: Adding the `byteorder` crate just for this one line seems to be overkill. It's probably better to leave out the dependency and use the standard library to do this:

```rust
for (i, &byte) in dev_id.to_be_bytes().iter().enumerate() {
    match dev_id_buffer.as_mut().get_mut(i) {
        None => return ReturnCode::ENOMEM,
        Some(value) => *value = byte,
    }
}
```

kernel/h1_syscalls/src/fuse.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    fn allow(&self,
+             app_id: AppId,
+             minor_num: usize,
+             slice: Option<AppSlice<Shared, u8>>
+    ) -> ReturnCode {
+        //debug!("allow: num={}, slice={}",
+        //    minor_num, if slice.is_some() { "Some" } else { "None" });
```
jrvanwhy: Probably best to remove this debugging code.

kernel/h1/src/fuse.rs
osenft: Fair :-)

kernel/h1_syscalls/src/fuse.rs
osenft: Done.

kernel/h1_syscalls/src/fuse.rs
osenft: I didn't actually know that there's support in the standard library for that, incl. `from_be_bytes`.  Thank you! Changed.

dann21 (approved):
Don't really have that much of the big picture in my head, but read through it and seems reasonable.

LGTM.

kernel/h1_syscalls/src/fuse.rs
jrvanwhy: Did you forget to commit the change? I still see it.

osenft: Argh. I did remove that. No idea where that got lost. Added (again) now.

jrvanwhy (approved):

PR #210, opened 2021-04-07
otpilot: Fixed handling of errors during SPI processing.

osenft:
We were previously not clearing the buffer, which resulted in
the message being processed repeatedly.

### DIFF
--- a/userspace/otpilot/src/main.rs
+++ b/userspace/otpilot/src/main.rs
@@ -163,6 +163,8 @@ fn run() -> TockResult<()> {
                             // Ignore error from writeln. There's nothing we can do here anyway.
                             let _ = writeln!(console, "Device: Error ending transaction.");
                         }
+                    } else {
+                        spi_device::get().end_transaction();
                     }
                 }
             }
--- a/userspace/otpilot/src/spi_device.rs
+++ b/userspace/otpilot/src/spi_device.rs
@@ -43,6 +43,9 @@ pub trait SpiDevice {
     /// Whether the transaction has the WRITE ENABLE bit set.
     fn is_write_enable_set(&self) -> bool;
 
+    /// End the transaction without changing any status bits or returning data.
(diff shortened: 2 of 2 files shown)


userspace/otpilot/src/main.rs
```diff
@@ -163,6 +163,8 @@ fn run() -> TockResult<()> {
                             // Ignore error from writeln. There's nothing we can do here anyway.
                             let _ = writeln!(console, "Device: Error ending transaction.");
                         }
+                    } else {
+                        spi_device::get().end_transaction();
```
dann21, line 167: Should `spi_device::get()` be pulled out into a local variable before line 155?  Will that simplify the code a bit?

dann21: `BUSY` is not set here, so it doesn't need to be cleared.  It looks like this end-transaction path ignores `WRITE_ENABLE`.  Is it a problem that this path leaves `WRITE_ENABLE` set?  (I have no idea.)

dann21 (approved):

userspace/otpilot/src/main.rs
osenft: > Should `spi_device::get()` be pulled out into a local variable before line 155? Will that simplify the code a bit?

`spi_device` is a static object, so I decided to always explicitly call the getter to don't give a false impression about it being "just an instance".

osenft: > `BUSY` is not set here, so it doesn't need to be cleared. It looks like this end-transaction path ignores `WRITE_ENABLE`. Is it a problem that this path leaves `WRITE_ENABLE` set? (I have no idea.)

We HAVE to clear `BUSY` here to allow the next transaction from coming in. However, the semantics for `WRITE_ENABLE` cannot be determined at this point (since an error occurred), so we leave it alone in both cases.

PR #211, opened 2021-04-09
otpilot: Fix main loop to not use console before yieldk.

osenft:
We used to print "yieldk()" to the console after checking conditions
but before actually calling yieldk() in the main loop. Since printing
to the console may result in calls to yieldk(), this resulted in the
main loop missing changed conditions and going into yieldk() again,
which could result in intermittent hangs.

### DIFF
--- a/userspace/otpilot/src/main.rs
+++ b/userspace/otpilot/src/main.rs
@@ -147,7 +147,9 @@ fn run() -> TockResult<()> {
 
     loop {
         while !spi_device::get().have_transaction() {
-            let _ = writeln!(console, "yieldk()");
+
+            // Note: Do NOT use the console here, as that results in a "hidden"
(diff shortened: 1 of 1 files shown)


tjshep (approved):
lgtm.    And I've been running the test_mailbox test in a loop this morning on the auto build results from this pr and it is not wedging up like it was sporadically wedging up before.   So this seems to have fixed the bug.

Good comment.   I do wonder if there are other places where writing to the console could cause similar problems.

osenft:
> I do wonder if there are other places where writing to the console could cause similar problems.

The problem is caused by us checking the conditions first, then printing and then `yieldk`-ing. If we check and then `yieldk`, it's fine. So no, there shouldn't be any other places in the code, since we're not checking the conditions anywhere else.

PR #212, opened 2021-04-09
otpilot: Fixed handling of errors during SPI processing.

osenft:
We were previously not clearing the buffer, which resulted in the message causing the error being processed repeatedly.


osenft:
Duplicate PR. Sorry!

PR #213, opened 2021-04-09
otpilot: Make spi_processor less noisy

osenft:
This change adds explicit handling for SPI command 0xC8 (read extended address register). It further adds an option for conditionally enabling the SpiProcessor to print the headers of incoming SPI packets, defaulting to "off" (it was previously always printing these headers, causing a lot of noise on the console).

### DIFF
--- a/userspace/otpilot/src/spi_processor.rs
+++ b/userspace/otpilot/src/spi_processor.rs
@@ -88,7 +88,11 @@ impl From<core::fmt::Error> for SpiProcessorError {
 //////////////////////////////////////////////////////////////////////////////
 
 pub struct SpiProcessor<'a> {
+    // The Manticore protocol server.
     pub server: PaRot<'a, Identity, Reset, NoRsa>,
+
+    // Whether to print incoming flash headers.
+    pub print_flash_headers: bool,
 }
 
 const SPI_TX_BUF_SIZE : usize = 512;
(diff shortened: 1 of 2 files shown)


userspace/otpilot/src/spi_processor.rs
```diff
@@ -269,6 +283,11 @@ impl<'a> SpiProcessor<'a> {
                 }
                 self.clear_device_status(true, true)
             }
+            OpCode::ReadExtendedAddressRegister => {
+                // We can't handle this, but we also don't want it to raise an
+                // UnsupportOpCode error.
+                self.ignore_transaction()
```
dann21: Would it make sense to spew something here, so that an unsuspecting victim will see evidence that the command they sent is getting ignored?

dann21 (approved):
LGTM (modulo one tentative suggestion)

userspace/otpilot/src/spi_processor.rs
osenft: You could make the argument that we generally want to print information about unsupported commands. I found this particular one by testing against u-boot running on Nuvoton. I'm not why it's making this call since the extended address register is not supported in Haven. Looks like a bug in the Nuvoton u-boot code to me, really. Since it's not supported, it'll return `0xff` on SPI. I believe that works exactly the same as if this was attempted with a regular SPI flash chip that doesn't support the command.

osenft: With that, I think you're right, we should not just silently ignore it. I removed the special handling code for the `ReadExtendedAddressRegister` again. This results in `process_spi_header()` returning an error which is then handled (by printing) in the `run()` loop. I think this is actually cleaner.

dann21 (approved):

PR #214, opened 2021-04-12
otpilot: Add Fuse controller.

osenft:
Note that this PR functionally depends on #209.

### DIFF
--- a/userspace/otpilot/src/fuse.rs
+++ b/userspace/otpilot/src/fuse.rs
@@ -14,7 +14,6 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-use byteorder::ByteOrder;
 use core::mem;
 use libtock::result::TockResult;
 use libtock::syscalls;
@@ -68,7 +67,7 @@ impl FuseImpl {
 
 impl Fuse for FuseImpl {
     fn get_dev_id(&self) -> TockResult<u64> {
-        let mut dev_id_buffer: [u8; mem::size_of::<u64>()] = [0; mem::size_of::<u64>()];
+        let mut dev_id_buffer = [0u8; mem::size_of::<u64>()];
 
         {
             // We want this to go out of scope after executing the command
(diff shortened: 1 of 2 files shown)


jrvanwhy (approved):

userspace/otpilot/src/fuse.rs
```diff
@@ -0,0 +1,82 @@
...
+    fn initialize(&'static mut self) -> TockResult<()> {
+        syscalls::command(DRIVER_NUMBER, command_nr::CHECK_IF_PRESENT, 0, 0)?;
+
+        Ok(())
+    }
+}
+
+impl Fuse for FuseImpl {
+    fn get_dev_id(&self) -> TockResult<u64> {
+        let mut dev_id_buffer: [u8; mem::size_of::<u64>()] = [0; mem::size_of::<u64>()];
```
jrvanwhy: I approved, then spotted something on a second pass. I don't think you need to specify the type here, and I think this would be clear enough:

```rust
let mut dev_id_buffer = [0u8; mem::size_of::<u64>()];
```

userspace/otpilot/src/fuse.rs
```diff
@@ -0,0 +1,82 @@
...
+
+        {
+            // We want this to go out of scope after executing the command
+            let _dev_id_buffer_share = syscalls::allow(DRIVER_NUMBER, allow_nr::DEV_ID_BUFFER, &mut dev_id_buffer)?;
+
+            syscalls::command(DRIVER_NUMBER, command_nr::GET_DEV_ID, 0, 0)?;
+        }
+
+        Ok(byteorder::BE::read_u64(&dev_id_buffer))
```
jrvanwhy: Should probably use the standard library directly rather than `byteorder`:

```rust
Ok(u64::from_be_bytes(dev_id_buffer))
```

userspace/otpilot/src/fuse.rs
osenft: Fixed (in second commit in this PR).

userspace/otpilot/src/fuse.rs
osenft: Oh yeah, I'm sorry. I should have updated that.

jrvanwhy (approved):

PR #215, opened 2021-04-12
otpilot: Add console_reader.

osenft:
The console_reader is a prerequisite for handling console input for debugging and testing.


dann21 (commented):
This is (obviously) not thread-safe.

Is that OK?

jrvanwhy (approved):

jrvanwhy:
> This is (obviously) not thread-safe.
> 
> Is that OK?

Yes, `libtock-rs` provides a purely single-threaded environment. A future version of `libtock-rs` will provide tools to make writing code like this safe.

PR #217, opened 2021-04-14
Make builds under `/tmp` work.

jrvanwhy:
Previously, the sandbox caused builds to fail (with a very confusing error message) if you checked out tock-on-titan under `/tmp`. The problem is the mount of `/tmp` in the sandbox hid the source code. With this change, `/tmp` and the source code are both visible and usable.

This change has no effect on builds outside `/tmp`.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
dfe90632efff924c82c1a8af40fe5ebd2097db58
git status
On branch tmp-build-fix
Your branch is up to date with 'origin/tmp-build-fix'.

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,12 @@ BUILD_SUBDIRS := kernel runner third_party tools userspace
 # installing things during builds. We don't like that. This is a sandbox we can
 # run commands in that denies network access as well as write access outside the
 # build/ directory.
+#
+# The mount order of /tmp and CURDIR is important. It is reasonable for someone
+# to check this repository out under /tmp and try to build it. If this only has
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -27,6 +27,8 @@ BUILD_SUBDIRS := kernel runner third_party tools userspace
 # build/ directory.
 BWRAP := bwrap                                                               \
          --ro-bind / /                                                       \
+         --tmpfs /tmp                                                        \
```
osenft, line 36: Can you please add a comment about this?

jrvanwhy: Good point, done.

osenft (approved):

tjshep (approved):
LGTM

Tested under /tmp and not under /tmp, also tested running "make build" from down in userspace/otpilot .    All seems good.

PR #220, opened 2021-04-27
otpilot: Use dynamic SPI device AddressConfig.

osenft:
This change prepares the otpilot code to enable dynamic configuration
of the SPI device addresses depending on the runtime configuration.

### DIFF
--- a/shared-lib/spiutils/src/driver.rs
+++ b/shared-lib/spiutils/src/driver.rs
@@ -25,6 +25,7 @@ use crate::protocol::wire::ToWire;
 
 use core::convert::TryFrom;
 use core::default::Default;
+use core::mem;
 
 /// Handler mode.
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
@@ -58,6 +59,9 @@ impl TryFrom<usize> for HandlerMode {
         }
     }
 }
+
+/// The length of an AddressConfig on the wire, in bytes.
+pub const ADDRESS_CONFIG_LEN: usize = 5 * mem::size_of::<u32>();
 
 /// Address configuration for SPI device hardware.
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
--- a/userspace/otpilot/src/spi_device.rs
+++ b/userspace/otpilot/src/spi_device.rs
@@ -16,14 +16,19 @@
 
 use core::cell::Cell;
 use core::convert::TryFrom;
+use core::mem;
 
 use libtock::result::TockError;
 use libtock::result::TockResult;
 use libtock::shared_memory::SharedMemory;
 use libtock::syscalls;
 
+use spiutils::driver::AddressConfig;
+use spiutils::driver::ADDRESS_CONFIG_LEN;
 use spiutils::driver::HandlerMode;
+use spiutils::io::Cursor;
 use spiutils::protocol::flash::AddressMode;
+use spiutils::protocol::wire::ToWire;
 
 pub const MAX_READ_BUFFER_SIZE: usize = 512;
 
@@ -68,6 +73,9 @@ pub trait SpiDevice {
 
     /// Set the SFDP data.
     fn set_sfdp(&self, data: &mut[u8]) -> TockResult<()>;
+
+    /// Configure SPI addresses.
+    fn configure_addresses(&self, address_config: AddressConfig) -> TockResult<()>;
 }
 
 // Get the static SpiDevice object.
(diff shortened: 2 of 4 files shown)


shared-lib/spiutils/src/driver.rs
```diff
@@ -59,6 +60,9 @@ impl TryFrom<usize> for HandlerMode {
     }
 }
 
+/// The length of an AddressConfig on the wire, in bytes.
+pub const ADDRESS_CONFIG_LEN: usize = 5 * mem::size_of::<u32>();
```
dann21, line 64: Hmm.  Shouldn't this constant be defined in the place that the serialization/deserialization code lives?

userspace/otpilot/src/spi_device.rs
```diff
@@ -280,4 +289,24 @@ impl SpiDevice for SpiDeviceImpl {
...
+        {
+            // Scope for cursor (which doesn't implement Drop).
+            // We need cursor to go out of scope so that we can use buf further down.
+            let cursor = Cursor::new(&mut buf);
+            if address_config.to_wire(cursor).is_err() {
+                return Err(TockError::Format);
+            }
+        }
+
+        // We want this to go out of scope after executing the command
```
dann21: This comment seems gratuitous.  Is it left over from some other formulation, in which the associated statement had to have its own scope>

userspace/otpilot/src/spi_processor.rs
```diff
@@ -43,6 +43,18 @@ use spiutils::protocol::wire::FromWireError;
 use spiutils::protocol::wire::ToWire;
 use spiutils::protocol::wire::ToWireError;
 
+// Size of the SPI flash chip.
+// Hard-coded to 64 MiB for now.
+// TODO: Modify this to be read from the actual SPI flash chip at runtime.
```
dann21: For 25+ years now, I've grown used to a convention wherein all TODO and FIXME comments are signed and dated.  This makes it easy for the reader to know useful information about them, along with having the very desirable side effect of embarrassing people who leave a large number of them around.
So I would write this as: `// TODO [osk 2021-04-27]: Modify this to be...`.

userspace/otpilot/src/spi_processor.rs
```diff
@@ -234,13 +246,13 @@ impl<'a> SpiProcessor<'a> {
         match header.opcode {
             OpCode::PageProgram => {
                 match header.get_address() {
-                    Some(0x02000000) => {
+                    Some(x) if x >= SPI_MAILBOX_ADDRESS && x < SPI_MAILBOX_ADDRESS + SPI_MAILBOX_SIZE => {
```
dann21: Would it make sense to factor `x >= SPI_MAILBOX_ADDRESS && x < SPI_MAILBOX_ADDRESS + SPI_MAILBOX_SIZE` into its own function?  It looks like it (or its negation) is repeated in four places.

dann21 (approved):
A few nit-picky comments, but LGTM overall.

shared-lib/spiutils/src/driver.rs
osenft: The de-/serialization code is in lines 87-113. I've always had the length just above the struct (see other code in this directory).

userspace/otpilot/src/spi_device.rs
osenft: I think it's still correct. The point is that we're declaring a variable prefixed with `_` since it's unused. Removing the declaration would have the side effect of dropping the object immediately.

shared-lib/spiutils/src/driver.rs
dann21: Aieee!!

userspace/otpilot/src/spi_device.rs
dann21: OK, I missed that point.  Maybe add to the comment?  Something like:

    // Assign return value to a(n unused) variable to keep it alive while the command is being executed.

userspace/otpilot/src/spi_processor.rs
osenft: Yeah you're right. We actually usually create bugs/issues, since they're trackable. I'll need to sort that out on this repo. For now, I just added my username as per Google convention.

userspace/otpilot/src/spi_processor.rs
osenft: Well spotted! Fixed.

userspace/otpilot/src/spi_device.rs
osenft: Done.

jrvanwhy (approved):

PR #222, opened 2021-04-29
otpilot: Add alarm driver.

osenft:
The use case for this change is to be able to ignore GPIO (follow-up PR) events for a given amount of time.

### DIFF
--- a/userspace/otpilot/src/alarm.rs
+++ b/userspace/otpilot/src/alarm.rs
@@ -1,0 +1,151 @@
+// Copyright 2021 lowRISC contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


userspace/otpilot/src/alarm.rs
```diff
@@ -0,0 +1,151 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use core::cell::Cell;
+
+use libtock::result::TockResult;
+use libtock::syscalls;
+
+pub trait Alarm {
+    // Get clock frequency in Hz.
```
dann21, line 23: I assume that this is the "tick" frequency.  It may be worth it to say that somewhere.

dann21 (approved):
LGTM

jrvanwhy (approved):

PR #224, opened 2021-04-29
Rename spiutils::driver to spiutils::driver::spi_device.

osenft:
This allows for other "driver" related code to be included in the spitutils package.


dann21 (approved):
LGTM

PR #225, opened 2021-04-29
otpilot: Add GPIO handling.

osenft:
This change adds the GPIO driver as well as a GPIO processor that handles
BMC reset monitor events.

### DIFF
--- a/userspace/otpilot/src/gpio.rs
+++ b/userspace/otpilot/src/gpio.rs
@@ -120,24 +120,30 @@ fn get_impl() -> &'static GpioControlImpl {
 }
 
 
+/// Adds one event.
+/// If the maximum number of events would be exceeded, do nothing.
 fn add_event(events: &Cell<usize>) {
     let val = events.get();
-    events.set(val + 1);
+    if val < core::usize::MAX - 1 {
+        events.set(val + 1);
+    }
 }
 
+/// Consumes one event.
+/// Returns the number of events to consume before consuming one.
 fn delete_event(events: &Cell<usize>) -> usize {
     let val = events.get();
-    if val > 0 {
+    if val > core::usize::MIN {
         events.set(val - 1);
     }
     val
 }
 
+/// Clears all events.
+/// Returns the number of events to consume before clearing them.
 fn clear_event(events: &Cell<usize>) -> usize {
     let val = events.get();
-    if val > 0 {
-        events.set(0);
-    }
+    events.set(core::usize::MIN);
     val
 }
 
--- a/userspace/otpilot/src/gpio.rs
+++ b/userspace/otpilot/src/gpio.rs
@@ -1,0 +1,217 @@
+// Copyright 2021 lowRISC contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use core::cell::Cell;
+use core::convert::TryFrom;
+
+use libtock::gpio::GpioPinUnitialized;
+use libtock::gpio::GpioPinRead;
+use libtock::gpio::GpioPinWrite;
+use libtock::gpio::{IrqMode, InputMode};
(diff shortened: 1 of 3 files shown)


userspace/otpilot/src/main.rs
```diff
@@ -148,9 +154,19 @@ fn run() -> TockResult<()> {
 
     //////////////////////////////////////////////////////////////////////////////
 
+    // We assume that we've already done all boot-time checks at this point.
```
dann21, line 157: Is there any way to assert this?

userspace/otpilot/src/main.rs
```diff
@@ -148,9 +154,19 @@ fn run() -> TockResult<()> {
...
+
+    // Deassert BMC resets.
+    // TODO(osk): Do something with the result codes.
```
dann21, line 160: "For now", maybe assert success?

userspace/otpilot/src/main.rs
```diff
@@ -183,9 +199,24 @@ fn run() -> TockResult<()> {
             console_reader::get().allow_read(1)?;
         }
 
+        if gpio::get().have_events() {
+            match gpio_processor.process_gpio_events() {
+                Ok(()) => {}
+                Err(_) => {
+                    // Ignore error from writeln. There's nothing we can do here anyway.
+                    let _ = writeln!(console, "Device: Error processing GPIO event.");
```
dann21, line 207: Is it possible to add any detail about the event for which processing failed?

userspace/otpilot/src/main.rs
```diff
@@ -183,9 +199,24 @@ fn run() -> TockResult<()> {
...
+        }
+
         if alarm::get().is_expired() {
-            writeln!(console, "Alarm expired.")?;
-            alarm::get().clear()?;
+            match gpio_processor.alarm_expired() {
+                Ok(()) => {}
+                Err(_) => {
+                    // Ignore error from writeln. There's nothing we can do here anyway.
+                    let _ = writeln!(console, "Alarm: Error processing event.");
```
dann21, line 217: "Error processing event."?  Is that right?

userspace/otpilot/src/gpio.rs
```diff
@@ -0,0 +1,217 @@
...
+            IS_INITIALIZED = true;
+        }
+        &GPIO_CTRL
+    }
+}
+
+
+fn add_event(events: &Cell<usize>) {
+    let val = events.get();
+    events.set(val + 1);
```
dann21: No worry about wraparound here?

userspace/otpilot/src/gpio.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+fn delete_event(events: &Cell<usize>) -> usize {
+    let val = events.get();
+    if val > 0 {
```
dann21: Hmm.  Should it be OK to silently do nothing here when `val == 0`?

userspace/otpilot/src/gpio.rs
```diff
@@ -0,0 +1,217 @@
...
+        events.set(val - 1);
+    }
+    val
+}
+
+fn clear_event(events: &Cell<usize>) -> usize {
+    let val = events.get();
+    if val > 0 {
```
dann21: Why conditionalize?  Why not always call `events.set(0)` here?

"Conditionals are the source of all complexity" -- rhh, ca. 1983

dann21 (approved):
LGTM.
Comments are a bunch of nit-picks.

userspace/otpilot/src/main.rs
osenft: This is more of a "design it in such a way that ..." type of comment.

userspace/otpilot/src/main.rs
osenft: We need to implement proper "boot time checks" during which we'd want to handle errors like that better, e.g. by resetting Haven or so. I think of this section as a placeholder to implement something.

userspace/otpilot/src/main.rs
osenft: I created #227 to generally implement better handling in these places.

userspace/otpilot/src/main.rs
osenft: Yeah - it's an alarm event. I think that's ok.

userspace/otpilot/src/gpio.rs
osenft: Good call! Fixed.

userspace/otpilot/src/gpio.rs
osenft: I added a comment explaining what this method is doing. So yes, that's ok.

userspace/otpilot/src/gpio.rs
osenft: Fair point. Removed.

PR #226, opened 2021-05-01
userspace: Fix build time detection of images that are too large.

osenft:
We actually support apps that are up to 64 KiB in size. Also, the
existing check didn't work because `wc` was not producing output
in the expected format.


dann21 (approved):
LGTM

PR #229, opened 2021-05-04
otpilot: Add console_processor.

osenft:
This change adds a simple interface to allow for basic manual operation of otpilot via serial console, e.g. for development and debugging.


dann21 (approved):
LGTM

PR #228, opened 2021-05-04
spiutils: Add firmware protocol.

osenft:
This change adds the protocol to update firmware.

PR #230, opened 2021-05-04
spiutils: Add SegmentInfo, RuntimeSegmentInfo and BuildInfo.

osenft:
This PR depends on #228

### DIFF
--- a/shared-lib/spiutils/src/io.rs
+++ b/shared-lib/spiutils/src/io.rs
@@ -243,6 +243,58 @@ impl BeInt for u64 {
     }
 }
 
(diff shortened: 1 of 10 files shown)


shared-lib/spiutils/src/io.rs
```diff
@@ -243,6 +243,58 @@ impl BeInt for u64 {
     }
 }
 
+/// A big-endian integer, which can be read and written.
```
dann21, line 246: "big"-endian?

dann21 (approved):
LGTM

PR #231, opened 2021-05-04
kernel: Add GlobalSec.

osenft:
This adds explicit support for H1's Global Security Module,
in particular to automatically configure flash protection
for the inactive RO and RW.

### DIFF
--- a/kernel/h1/src/globalsec.rs
+++ b/kernel/h1/src/globalsec.rs
@@ -1,0 +1,220 @@
+// Copyright 2021 lowRISC contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::hil::globalsec::GlobalSec;
+
+use kernel::common::registers::register_bitfields;
+use kernel::common::registers::register_structs;
+use kernel::common::registers::ReadWrite;
+use kernel::common::StaticRef;
+
+use spiutils::driver::firmware::RuntimeSegmentInfo;
+use spiutils::driver::firmware::SegmentInfo;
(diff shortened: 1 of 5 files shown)


kernel/h1/src/globalsec.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+impl GlobalSecHardware {
+    const fn new(base_addr: StaticRef<Registers>) -> GlobalSecHardware {
+        let globalsec = GlobalSecHardware {
+            registers: base_addr,
+            runtime_segment_info: UNKNOWN_RUNTIME_SEGMENT_INFO,
+        };
+
+        globalsec
```
jrvanwhy: You don't need the `globalsec` local variable here, just return the new struct directly.

kernel/h1_syscalls/src/globalsec.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    fn command(&self, command_num: usize, _arg1: usize, _arg2: usize, caller_id: AppId)
+        -> ReturnCode {
+        if self.current_user.get() == None {
+            self.current_user.set(Some(caller_id));
+        }
+        match command_num {
+            0 /* Check if present */ => ReturnCode::SUCCESS,
+            1 /* Get runtime segmenet info */ => self.get_runtime_segment_info(caller_id),
```
jrvanwhy: segment is misspelled in the comment.

kernel/h1_syscalls/src/globalsec.rs
```diff
@@ -0,0 +1,104 @@
...
+        match minor_num {
+            0 => {
+                // Buffer for data exchange
+                self.apps
+                    .enter(app_id, |app_data, _| {
+                        if let Some(s) = slice {
+                            app_data.buffer = Some(s);
+                        } else {
+                            app_data.buffer = slice;
+                        }
```
jrvanwhy: Isn't this whole `if` statement just `app_data.buffer = slice;`?

kernel/h1/src/globalsec.rs
osenft: No idea what I was thinking there. Fixed.

kernel/h1_syscalls/src/globalsec.rs
osenft: I copied this pattern a few capsules ago from somewhere ... I think. Anyway. Good point. Fixed.

kernel/h1_syscalls/src/globalsec.rs
osenft: Argh. I kept making this same typo and though I had caught it everywhere. Thanks. Fixed.

jrvanwhy (approved):

PR #234, opened 2021-05-05
kernel: Add Reset component and capsule.

osenft:
This adds a HIL, implementation and capsule to execute a chip reset and to get the source of the last reset.

Depends on PR #237.

### DIFF
--- a/kernel/h1/src/pmu.rs
+++ b/kernel/h1/src/pmu.rs
@@ -41,8 +41,13 @@
 //!     * Designed for 1.8-3.6V
 //!
 
+use crate::hil::reset;
+
 use core::mem::transmute;
 use kernel::common::cells::VolatileCell;
+
+/// Value to write into global_reset register to initiate a reset.
+const GLOBAL_RESET_VALUE: u32 = 0x7041776;
 
 /// Registers for the Power Management Unit (PMU)
 // Non-public fields prefixed with "_" mark unused registers
@@ -52,7 +57,7 @@ pub struct PMURegisters {
     _set_reset: VolatileCell<u32>,
 
     /// Clear register for the reset source
-    pub clear_reset: VolatileCell<u32>,
+    clear_reset: VolatileCell<u32>,
 
     /// Status for source of last reset event
     ///
@@ -69,13 +74,13 @@ pub struct PMURegisters {
     /// | 6   | Fast burnout circuit                              |
     /// | 7   | Security breach reset                             |
     ///
-    pub reset_source: VolatileCell<u32>,
+    reset_source: VolatileCell<u32>,
 
     /// Global chip reset
     ///
     /// Initiates a reset of the system similar to toggling the external reset
     /// pin. To initiate a reset, write the key 0x7041776 to this register.
-    pub global_reset: VolatileCell<u32>,
+    global_reset: VolatileCell<u32>,
 
     pub low_power_disable: VolatileCell<u32>,
 
@@ -149,19 +154,21 @@ pub struct PMURegisters {
 
     pub _gate_on_sleep_set1: VolatileCell<u32>,
     pub _gate_on_sleep_clr1: VolatileCell<u32>,
-    
+
     pub _clock0: VolatileCell<u32>,
     pub _reset0_write_enable: VolatileCell<u32>,
     pub reset0: VolatileCell<u32>,
 
     pub _reset1_write_enable: VolatileCell<u32>,
     pub _reset1: VolatileCell<u32>
-    
+
 }
 
 const PMU_BASE: isize = 0x40000000;
 
-static mut PMU: *mut PMURegisters = PMU_BASE as *mut PMURegisters;
+pub static mut PMU: *mut PMURegisters = PMU_BASE as *mut PMURegisters;
+
+pub static mut RESET: Reset = Reset::new();
 
 #[derive(Clone,Copy)]
 pub enum PeripheralClock0 {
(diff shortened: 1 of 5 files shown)


kernel/h1/src/pmu.rs
```diff
@@ -41,9 +41,14 @@
 //!     * Designed for 1.8-3.6V
 //!
 
+use crate::hil::reset;
+
 use core::mem::transmute;
 use kernel::common::cells::VolatileCell;
 
+/// Value to write into global_reset register to initiate a reset.
+const GLOBAL_RESET_VALUE: u32 = 0x7041776;
```
dann21: Where does this value come from?  (Maybe add a reference to the comment?)

kernel/h1/src/pmu.rs
```diff
@@ -149,19 +154,21 @@ pub struct PMURegisters {
...
     pub _reset0_write_enable: VolatileCell<u32>,
     pub reset0: VolatileCell<u32>,
 
     pub _reset1_write_enable: VolatileCell<u32>,
     pub _reset1: VolatileCell<u32>
-    
+
 }
 
 const PMU_BASE: isize = 0x40000000;
```
dann21, line 170: Maybe add a comment here that this is the base-address of the PMU registers?
It probably seems obvious to you, but I did a double-take when I saw that this value was getting coerced into a pointer to a PMURegisters object (below).  "How can it be right to coerce an isize into a pointer?"

kernel/h1/src/pmu.rs
```diff
@@ -269,3 +276,45 @@ pub fn reset_dcrypto() {
...
+    pub fn init(&mut self) {
+        let pmu: &mut PMURegisters = unsafe { transmute(PMU) };
+
+        // Read and reset the reset source
+        self.reset_source = unsafe{(pmu.reset_source.get() & 0xff) as u8};
+        unsafe{pmu.clear_reset.set(1)};
+    }
+}
+
+impl reset::Reset for Reset {
```
dann21: This may just be that I don't know rust at all, but I find this line to be confusing.  It uses one name `[rR]eset` for three different things (value, object and trait, if I'm reading it correctly).  Is that standard practice?  (Yes, I know, it's rust.  If it was hard to write, it should be hard to read.  :-) )

kernel/h1/src/pmu.rs
```diff
@@ -269,3 +276,45 @@ pub fn reset_dcrypto() {
...
+
+impl Reset {
+    const fn new() -> Reset {
+        Reset {
+            reset_source: 0,
+        }
+    }
+
+    pub fn init(&mut self) {
+        let pmu: &mut PMURegisters = unsafe { transmute(PMU) };
```
dann21, line 296: It seems ugly to me that you need to do a `memcpy()` (== `transmute()`) to coerce from a `*mut PMURegisters` to a `&mut PMURegisters`.  Isn't there a more principled way to do that?

Once again, I don't know the language.

kernel/h1/src/pmu.rs
```diff
@@ -269,3 +276,45 @@ pub fn reset_dcrypto() {
...
+    fn reset(&self) -> ! {
+        let pmu: &mut PMURegisters = unsafe { transmute(PMU) };
+
+        unsafe {pmu.global_reset.set(GLOBAL_RESET_VALUE)};
+
+        // Wait for reboot; should never return
+        loop {
+            unsafe {
+                asm!("dsb" :::: "volatile");
+                asm!("wfi" :::: "volatile");
```
dann21, line 314: The casual reader (me) has no idea at all what these instructions are doing.

Furthermore, a loop that spins, executing two instructions one after another forever seems wrong.  Presumably that's not what it's doing, because presumably the sequence causes the machine to reset, but I had to infer that.  There should at least be a comment that says that.

Finally, if the two-instruction sequence should always cause a reset, then this is not the way I would write it -- I would write it as something like `dsb`; `wfi`; `UNREACHABLE()`.

Either that or I have no clue what's going on here.

kernel/h1_syscalls/src/reset.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub fn new(reset: &'a dyn Reset) -> ResetSyscall<'a> {
+        ResetSyscall {
+            reset: reset,
+        }
+    }
+
+    fn reset(&self) -> ReturnCode {
+        self.reset.reset();
+
+        // This should never return. If it did, something went wrong.
```
dann21: Is there no such thing as `UNREACHABLE()` to use here?

dann21 (commented):
I think I understand most of what's going on here, but I think I had a sufficient number of questions that I'll wait on approval until you make clear the depth of my cluelessness.

kernel/h1/src/pmu.rs
osenft: Yeah, there are better (I think) ways that I used for the implementation in other drivers, e.g. look at "fuse.rs".

However, the pmu.rs already existed and I didn't want to refactor the whole file. I just followed the style that was used in the existing code in this file.

kernel/h1/src/pmu.rs
osenft: I actually got that sequence from some of our existing code for this chip. I agree it's a little confusing, but I don't think it's exciting enough. It certainly does what we want it to do ...

kernel/h1/src/pmu.rs
osenft: Good point. Done.

kernel/h1/src/pmu.rs
osenft: Done.

kernel/h1/src/pmu.rs
osenft: There are actually 5 different "reset" things:
* struct (Reset)
* trait (Reset)
* namespace (reset)
* method (reset)
* instance (RESET)

I renamed it to make it a little clearer:
* struct (ResetImpl)
* trait (Reset)
* namespace (reset)
* method (reset_chip)
* instance (RESET)

kernel/h1_syscalls/src/reset.rs
osenft: I updated the comment. The compiler would flag it if `self.reset.reset_chip()` changed its return type and actually returned.

kernel/h1/src/pmu.rs
dann21: :+1:

kernel/h1/src/pmu.rs
dann21: ok

kernel/h1_syscalls/src/reset.rs
dann21: :+1:

kernel/h1/src/pmu.rs
dann21: I guess the comment above the loop gives enough of a high-level view of what's going on, although I'd love to know why it needs to be in a loop.  Perhaps it's written that way to convince the compiler that it never returns?

dann21 (approved):
LGTM, modulo whatever is causing the build failure.

PR #235, opened 2021-05-05
otpilot: Add reset driver and add to console.

osenft:
This adds support for the kernel's reset capsule to execute a chip reset and get the source for the last reset.

Depends on PR #237.

### DIFF
--- a/userspace/otpilot/src/reset.rs
+++ b/userspace/otpilot/src/reset.rs
@@ -1,0 +1,115 @@
+// Copyright 2021 lowRISC contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use libtock::result::TockResult;
+use libtock::syscalls;
+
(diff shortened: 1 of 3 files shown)


userspace/otpilot/src/reset.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn reset(&self) -> TockResult<()> {
+        syscalls::command(DRIVER_NUMBER, command_nr::RESET, 0, 0)?;
+
+        panic!("The Reset driver call should not have returned.")
+    }
+
+    fn get_reset_source(&self) -> TockResult<ResetSource> {
+        let reset_bits = syscalls::command(DRIVER_NUMBER, command_nr::GET_RESET_SOURCE, 0, 0)?;
+        Ok(ResetSource {
+            power_on_reset: (reset_bits & 0x1) != 0,
```
dann21: This can't be the only place these bits are defined -- they must be set somewhere.  I don't see where, but the two places should share definitions.

dann21 (approved):
LGTM, with one comment.

userspace/otpilot/src/reset.rs
osenft: I moved it into PR #237.

userspace/otpilot/src/spi_processor.rs
```diff
@@ -149,7 +149,7 @@ impl<'a> SpiProcessor<'a> {
     fn process_spi_payload(&mut self, mut data: &[u8]) -> SpiProcessorResult<()> {
         let mut console = Console::new();
         let header = payload::Header::from_wire(&mut data)?;
-        writeln!(console, "Device: payload header: {:?}", header)?;
+        //writeln!(console, "Device: payload header: {:?}", header)?;
```
dann21: Deleteme?

dann21 (approved):
LGTM

userspace/otpilot/src/spi_processor.rs
osenft: Ah, that shouldn't have been deleted here. I'll undo it by overwriting this last commit.

PR #237, opened 2021-05-05
spiutils: Add driver::reset::ResetSource.

osenft:
This change will be used to communicate the reset source from the kernel to userspace.


dann21 (approved):
LGTM

PR #240, opened 2021-05-12
otpilot: Consume SYS_RSTMON_N events.

osenft:
If we do not consume SYS_RSTMON_N events, we end up in an infinite
loop that never yields back into the kernel since there is always
an event to process.

### DIFF
--- a/userspace/otpilot/src/gpio_processor.rs
+++ b/userspace/otpilot/src/gpio_processor.rs
@@ -129,6 +129,11 @@ impl GpioProcessor {
             }
         }
 
+        let sys_rstmon_n = gpio::get().consume_event(GpioPin::SYS_RSTMON_N);
+        if sys_rstmon_n {
+            writeln!(console, "Ignored sys_rstmon_n")?;
+        }
+
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


userspace/otpilot/src/gpio_processor.rs
```diff
@@ -129,6 +129,11 @@ impl GpioProcessor {
             }
         }
 
+        let sys_rstmon_n = gpio::get().consume_event(GpioPin::SYS_RSTMON_N);
```
dann21, line 132: Hmm.  Are we certain that it's OK to ignore a reset event here?  Is it possible that this logic will cause us to ignore an intended reset?

(This is why the guy I learned most of what I know about distributed systems from _hates_ edges (i.e. events) -- he prefers levels.)

dann21 (approved):
LGTM, with a question.

userspace/otpilot/src/gpio_processor.rs
osenft: We don't have anything to do for SYS_RSTMON_N yet. You could argue that we might not want ANY handling for this GPIO at this time. However, I found it important to provide all the plumbing so that handling could be easily added later once we actually do attestation.

dann21: :+1:

PR #241, opened 2021-05-12
kernel/h1: Add support for both flash banks.

osenft:
This change adds support to the H1 flash driver for both flash banks.
In order to be able to use the existing interface / design, we now
need to pass the target address to both `H1bHw.set_transaction()` and
`H1bHw.trigger()`.

Further, this change adds `ReturnCode`s to the `Hardware` trait in
preparation to handle these in the driver code in a subsequent
commit.

### DIFF
--- a/kernel/h1/src/hil/flash/driver.rs
+++ b/kernel/h1/src/hil/flash/driver.rs
@@ -67,9 +67,11 @@ const MAX_WRITE_SIZE: usize = 32; // Maximum single write is 32 words
 impl<'d, A: Alarm<'d>, H: Hardware> super::flash::Flash<'d> for FlashImpl<'d, A, H> {
     fn erase(&self, page: usize) -> ReturnCode {
         if self.program_in_progress() { return ReturnCode::EBUSY; }
+        let target: usize = page * super::WORDS_PER_PAGE;
+        self.write_target.set(target);
         self.smart_program(ERASE_OPCODE, /*max_attempts*/ 45, /*final_pulse_needed*/ false,
                            /*timeout_nanoseconds*/ 3_353_267,
-                           /*target*/ page * super::WORDS_PER_PAGE, /*size*/ 1);
+                           /*target*/ target, /*size*/ 1);
         ReturnCode::SUCCESS
     }
 
@@ -80,7 +82,7 @@ impl<'d, A: Alarm<'d>, H: Hardware> super::flash::Flash<'d> for FlashImpl<'d, A,
     fn write(&self, target: usize, data: &'d mut [u32]) -> (ReturnCode, Option<&'d mut [u32]>) {
         let write_len = cmp::min(data.len(), MAX_WRITE_SIZE);
 
-        //if data.len() > 32 { return (ReturnCode::ESIZE, Some(data)); }
+        if data.len() > MAX_WRITE_SIZE { return (ReturnCode::ESIZE, Some(data)); }
         if self.program_in_progress() { return (ReturnCode::EBUSY, Some(data)); }
         self.write_pos.set(0);
         self.write_target.set(target);
@@ -110,7 +112,7 @@ impl<'d, A: Alarm<'d>, H: Hardware> ::kernel::hil::time::AlarmClient for FlashIm
     fn fired(&self) {
         if let Some(state) = self.smart_program_state.take() {
             let state = state.step(
-                self.alarm, self.hw, self.opcode.get());
+                self.alarm, self.hw, self.opcode.get(), self.write_target.get());
             if let Some(code) = state.return_code() {
                 if let Some(client) = self.client.get() {
                     if self.opcode.get() == WRITE_OPCODE {
@@ -161,7 +163,7 @@ impl<'d, A: Alarm<'d>, H: Hardware> FlashImpl<'d, A, H> {
         self.hw.set_transaction(target, size - 1);
         self.smart_program_state.set(Some(
             SmartProgramState::init(max_attempts, final_pulse_needed, timeout_nanoseconds)
-                .step(self.alarm, self.hw, opcode)));
+                .step(self.alarm, self.hw, opcode, target)));
         self.opcode.set(opcode);
     }
 }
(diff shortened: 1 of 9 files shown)


kernel/h1/src/hil/flash/h1_hw.rs
```diff
@@ -265,27 +266,56 @@ impl super::hardware::Hardware for H1bHw {
...
+        // Compute the offset relative to the flash bank.
+        let relative_offset: usize;
+        if offset < H1_FLASH_BANK_SIZE / BYTES_PER_WORD {
+            relative_offset = offset;
+        } else {
+            relative_offset = offset - H1_FLASH_BANK_SIZE / BYTES_PER_WORD;
+        }
+
+        if relative_offset + size > H1_FLASH_BANK_SIZE / BYTES_PER_WORD {
+            return ReturnCode::EINVAL
```
jrvanwhy: This is a lot of complexity for untestable code. Can some of this logic be moved out of `H1bHw`?

osenft: Good point. I moved that logic out into the `Driver`.

osenft: If you like I can squash the two commits and force push if that makes it easier to read.

kernel/h1/src/hil/flash/hardware.rs
```diff
@@ -14,8 +14,19 @@
 
 use kernel::ReturnCode;
 
-/// The interface between the flash driver and the (real or fake) flash module.
+/// The bank to perform an operation on.
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
+pub enum Bank {
+    Unknown,
+    Zero,
+    One,
```
jrvanwhy, line 21: Why not just:

```rust
pub enum Bank {
    Zero,
    One,
}
```

then use `Option<Bank>` (or a similarly-structured custom enum type) wherever you need a third value.

jrvanwhy: In particular, this would make `H1bHw::trigger` simpler.

kernel/h1/src/hil/flash/driver.rs
```diff
@@ -63,13 +63,26 @@ impl<'d, A: Alarm<'d>, H: Hardware> FlashImpl<'d, A, H> {
...
 const MAX_WRITE_SIZE: usize = 32; // Maximum single write is 32 words
+const WORDS_PER_BANK: usize = 0x10000; // 64ki words per bank
+
+// Computes the flash Bank for the specified target location in words
+// from the beginning of flash.
+fn get_bank_from_target(target: usize) -> super::hardware::Bank {
+    if target < WORDS_PER_BANK {
+        super::hardware::Bank::Zero
+    } else {
+        super::hardware::Bank::One
```
dann21: Should we assert `target < 2*WORDS_PER_BANK` here?

kernel/h1/src/hil/flash/hardware.rs
```diff
@@ -28,14 +39,15 @@ pub trait Hardware {
     fn read_error(&self) -> u16;
 
     /// Set flash transaction parameters (word offset and size). The word offset
-    /// is relative to the start of flash and the size is one less than the
+    /// is relative to the start of a flash bank and the size is one less than the
```
dann21, line 37: This `size-1` thing feels like a questionable choice to me.  It seems likely to lead to confusion in the future.

Why?

dann21 (approved):
LGTM, with a couple of comments.

kernel/h1/src/hil/flash/hardware.rs
osenft: The `Hardware` trait is really just a super thin layer above the actual HW registers. And the `-1` is just how it is in the actual hardware.

dann21:
Sigh...
OK, can't fix that.

kernel/h1/src/hil/flash/hardware.rs
dann21: Sigh...OK, can't fix that.

kernel/h1/src/hil/flash/hardware.rs
osenft: I tried to compromise wrt. required changes. But you're right, removing `Unknown` is absolutely better. Please see latest changes.

kernel/h1/src/hil/flash/driver.rs
osenft: The latest commits handle this correctly.

kernel/h1/src/hil/flash/driver.rs
```diff
@@ -63,13 +67,36 @@ impl<'d, A: Alarm<'d>, H: Hardware> FlashImpl<'d, A, H> {
...
+        if maybe_bank.is_none() {
+            return ReturnCode::EINVAL;
+        }
+
+        self.write_bank.set(maybe_bank.unwrap());
+        self.write_bank_target.set(target % WORDS_PER_BANK);
         self.smart_program(ERASE_OPCODE, /*max_attempts*/ 45, /*final_pulse_needed*/ false,
-                           /*timeout_nanoseconds*/ 3_353_267,
-                           /*target*/ page * super::WORDS_PER_PAGE, /*size*/ 1);
+                           /*timeout_nanoseconds*/ 3_353_267, self.write_bank.get(),
```
dann21, line 97: Where the _heck_ did this bizarre-looking magic number come from?

dann21 (approved):

kernel/h1/src/hil/flash/driver.rs
osenft: I assume @jrvanwhy computed that from the H1B specs ...

kernel/h1/src/hil/flash/driver.rs
dann21: Yep -- tx.

kernel/h1/src/hil/flash/driver.rs
dann21: Maybe not yours to dig up, but IMO anything like this ought to include a comment with a pointer to the relevant spec.

PR #242, opened 2021-05-12
kernel: Add capsule for the H1 flash driver.

osenft:


jrvanwhy (commented):
SGTM in principle, because we don't currently have a mechanism for identifying privileged apps. Eventually this will need to be exposed only to trusted apps.

Definitely some weird patterns, like the:

```rust
if let Some(s) = slice {
    app_data.write_buffer = Some(s);
} else {
    app_data.write_buffer = slice;
}
```

which is equivalent to `app_data.write_buffer = slice;`

osenft:
> Definitely some weird patterns, like the:
> 
> ```rust
> if let Some(s) = slice {
>     app_data.write_buffer = Some(s);
> } else {
>     app_data.write_buffer = slice;
> }
> ```
> 
> which is equivalent to `app_data.write_buffer = slice;`

Argh, yeah, I'm sorry. We talked about that before. Fixed.

jrvanwhy (approved):

dann21 (commented):
Looks OK to me, but I don't feel that I follow rust details well enough to approve.

PR #243, opened 2021-05-12
otpilot: Add flash support.

osenft:
This adds the userspace driver for H1's flash capsule.

### DIFF
--- a/userspace/otpilot/src/flash.rs
+++ b/userspace/otpilot/src/flash.rs
@@ -1,0 +1,174 @@
+// Copyright 2021 lowRISC contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use core::cell::Cell;
+
+use libtock::result::TockResult;
+use libtock::syscalls;
+use libtock::syscalls::raw::yieldk;
+
+pub const MAX_BUFFER_LENGTH: usize = 128;
+
+pub trait Flash {
+    // Read from flash.
+    // offset: Location relative to flash start to read from. Must be word (4 bytes) aligned.
+    // buffer: Buffer for read data.
+    //         It must be: buffer.len() <= len <= MAX_BUFFER_LENGTH
+    // read_length: Number of bytes to read. Must be > 0 and a multiple of 4.
+    fn read(&self, offset: usize, buffer: &mut[u8], len: usize) -> TockResult<()>;
+
+    // Write to flash.
+    // offset: Location relative to flash start to write to. Must be word (4 bytes) aligned.
+    // buffer: Buffer with data to write.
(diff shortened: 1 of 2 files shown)


userspace/otpilot/src/flash.rs
```diff
@@ -0,0 +1,174 @@
...
+use libtock::syscalls::raw::yieldk;
+
+pub const MAX_BUFFER_LENGTH: usize = 128;
+
+pub trait Flash {
+    // Read from flash.
+    // offset: Location relative to flash start to read from. Must be word (4 bytes) aligned.
+    // buffer: Buffer for read data.
+    //         It must be: buffer.len() <= len <= MAX_BUFFER_LENGTH
+    // read_length: Number of bytes to read. Must be > 0 and a multiple of 4.
```
dann21: Nit: comment (`read_length`) is not consistent with parameter name (`len`).

userspace/otpilot/src/flash.rs
```diff
@@ -0,0 +1,174 @@
...
+use libtock::syscalls;
+use libtock::syscalls::raw::yieldk;
+
+pub const MAX_BUFFER_LENGTH: usize = 128;
+
+pub trait Flash {
+    // Read from flash.
+    // offset: Location relative to flash start to read from. Must be word (4 bytes) aligned.
+    // buffer: Buffer for read data.
+    //         It must be: buffer.len() <= len <= MAX_BUFFER_LENGTH
```
dann21: Shouldn't `len` be `<= buffer.len()`?

userspace/otpilot/src/flash.rs
```diff
@@ -0,0 +1,174 @@
...
+    // read_length: Number of bytes to read. Must be > 0 and a multiple of 4.
+    fn read(&self, offset: usize, buffer: &mut[u8], len: usize) -> TockResult<()>;
+
+    // Write to flash.
+    // offset: Location relative to flash start to write to. Must be word (4 bytes) aligned.
+    // buffer: Buffer with data to write.
+    //         It must be: buffer.len() <= len <= MAX_BUFFER_LENGTH
+    // read_length: Number of bytes to write. Must be > 0 and a multiple of 4.
```
dann21: This should be `write_length`, not `read_length`.  And, again, the parameter name (`len`) is inconsistent with this comment.

userspace/otpilot/src/flash.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    fn clear_operation(&self);
+}
+
+// Get the static Flash object.
+pub fn get() -> &'static dyn Flash {
+    get_impl()
+}
+
+const DRIVER_NUMBER: usize = 0x40040;
```
dann21, line 61: Why does this magic number have this value?  Should there be a reference (comment) here?

dann21 (approved):
LGTM, as far as I understand it.

userspace/otpilot/src/flash.rs
osenft: I think this is just a bit of "tribal" Tock knowledge on where to look for this number. One would usually look for the board-specific implementation of Tock's `kernel::platform::Platform`, which for us is in `tock-on-titan/kernel/papa/src/main.rs`. This lists all the drivers supported for the particular board. From there you can find the magic numbers.

userspace/otpilot/src/flash.rs
osenft: Oh, good catch, thank you! Fixed.

userspace/otpilot/src/flash.rs
osenft: Oops ... Yeah, of course. Thanks! Fixed.

userspace/otpilot/src/flash.rs
osenft: Fixed.

PR #244, opened 2021-05-14
kernel/papa: Add flash syscalls.

osenft:
This change also removes the `nvcounter` and `personality` syscalls as they're not compatible with the `flash` syscall.


dann21 (approved):
LGTM

PR #245, opened 2021-05-14
kernel: Disable SPI device address filtering.

osenft:
The address filtering gets in the way if the BMC reads from addresses
from "outside" of the SPI flash chip's address range. This is the
case for BMCs that request read from an address with all high bits
set to 1.


dann21 (approved):
LGTM

PR #246, opened 2021-05-14
userspace: Generate gitlongtag without trailing newline.

osenft:
This change allows apps to consume the tag without having to strip
newline characters themselves.


dann21 (approved):
LGTM

PR #248, opened 2021-05-14
userspace: Enable build of both A and B images.

osenft:
This change adds the option for Rust userspace apps to be built with
both and A and a B image with different address offsets.

It also enables this functionality for the otpilot app.

### DIFF
--- a/kernel/Build.mk
+++ b/kernel/Build.mk
@@ -30,7 +30,7 @@ kernel/build$(IMAGE): sandbox_setup
 .PHONY: kernel/build-signed$(IMAGE)
 kernel/build-signed$(IMAGE): kernel/build$(IMAGE)
 
-endef
+endef # KERNEL_IMAGE_TARGETS
 
 # Instantiate for IMAGE=-a and IMAGE=-b
 $(foreach IMAGE,$(IMAGES),$(eval $(KERNEL_IMAGE_TARGETS)))
--- a/kernel/Build.mk
+++ b/kernel/Build.mk
@@ -12,12 +12,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: kernel/build
-kernel/build: sandbox_setup
-	cd kernel && $(BWRAP) cargo build --release
+IMAGES=_a _b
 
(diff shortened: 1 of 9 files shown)


kernel/Build.mk
```diff
@@ -12,12 +12,31 @@
...
+kernel/build$(IMAGE): sandbox_setup
+	cd kernel && \
+		CARGO_TARGET_DIR="../build/kernel/cargo$(IMAGE)" \
+		RUSTFLAGS="-C link-arg=-T./layout$(IMAGE).ld" \
+		$(BWRAP) cargo build --release
+
+.PHONY: kernel/build-signed$(IMAGE)
+kernel/build-signed$(IMAGE): kernel/build$(IMAGE)
+
+endef
```
dann21: Pernaps ` endef # KERNEL_IMAGE_TARGETS` would make this more readable?

userspace/Build.mk
```diff
@@ -279,42 +271,68 @@ userspace/$(APP)/$(BOARD)/doc: sandbox_setup build/gitlongtag
...
-userspace/$(APP)/$(BOARD)/program: \
+endef
+
+# ------------------------------------------------------------------------------
+# Macro to define targets for a specific board-app-and-image combination.
+# Arguments:
+# - $(BOARD)
+# - $(APP)
+# - $(IMAGE) [optional suffix to build A/B images]
+define RUST_APP_BOARD_IMAGE_TARGETS
```
dann21, line 282: IMO there ought to be a comment attached to the closing `endef` for this `define` that makes it easy for the reader to visually discern the entirety of the macro body.

dann21 (approved):
LGTM -- a couple of comment nits, that's all.

kernel/Build.mk
osenft: Good call. Done throughout.

userspace/Build.mk
osenft: Done throughout.

dann21 (approved):
LGTM (modulo build failure)

jrvanwhy (approved):

PR #250, opened 2021-05-14
spiutils: Add InactiveSegmentsInfo and RebootRequest/Response messages.

osenft:


shared-lib/spiutils/src/protocol/firmware.rs
```diff
@@ -322,3 +437,104 @@ impl ToWire for WriteChunkResponse {
         Ok(())
     }
 }
+
+// ----------------------------------------------------------------------------
+
+wire_enum! {
+    /// When to perform the reboot.
+    pub enum RebootTime: u8 {
```
dann21, line 445: Similarly to Johnathan's earlier comment, would this be better phrased as an Option?

shared-lib/spiutils/src/protocol/firmware.rs
```diff
@@ -322,3 +437,104 @@ impl ToWire for WriteChunkResponse {
...
+        w.write_be(self.time.to_wire_value())?;
+        Ok(())
+    }
+}
+
+// ----------------------------------------------------------------------------
+
+wire_enum! {
+    /// The result of a reboot request.
+    pub enum RebootResult: u8 {
```
dann21, line 489: Maybe an Option?

dann21 (approved):
LGTM, modulo a couple of comments about maybe using an Option instead of an enum that has an Unknown value.

shared-lib/spiutils/src/protocol/firmware.rs
osenft: Actually, we don't need or want the `Unknown` value. I removed it. Good catch!

shared-lib/spiutils/src/protocol/firmware.rs
osenft: Actually, we don't need or want the `Unknown` value. I removed it. Good catch!

dann21 (approved):

PR #251, opened 2021-05-15
spiutils + otpilot: Add checksum field and the Error message type.

osenft:
This change adds a checksum field to the
spiutils::protocol::payload::Header alongside code to compute the
checksum. It further adds the Error message type to report
checksum and payload content type errors back to the sender of
the message. The new checksum field and Error message types are
also added to otpilot's spi_processor in this change.


dann21 (approved):
LGTM

PR #252, opened 2021-05-15
spiutils: Remove "Unknown" values from wire_enums.

osenft:
We always want and need "known" values in these messages, so having
the "Unknwown" values is misleading.

### DIFF
--- a/shared-lib/spiutils/src/protocol/firmware.rs
+++ b/shared-lib/spiutils/src/protocol/firmware.rs
@@ -31,9 +31,6 @@ use crate::protocol::wire::WireEnum;
 wire_enum! {
     /// The content type.
     pub enum ContentType: u8 {
-        /// Unknown message type.
-        Unknown = 0xff,
-
         /// Request to prepare for an update
         UpdatePrepareRequest = 0x01,
 
@@ -262,9 +259,6 @@ impl ToWire for UpdatePrepareRequest {
 wire_enum! {
     /// The result of an update prepare request.
     pub enum UpdatePrepareResult: u8 {
-        /// Unknown result type.
-        Unknown = 0xff,
-
         /// Success
         Success = 0x00,
 
@@ -371,9 +365,6 @@ impl ToWire for WriteChunkRequest<'_> {
 wire_enum! {
     /// The result of a write chunk request.
     pub enum WriteChunkResult: u8 {
-        /// Unknown result type.
-        Unknown = 0xff,
-
         /// Success
         Success = 0x00,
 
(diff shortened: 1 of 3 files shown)


dann21 (approved):
LGTM, but added a few comments after.

shared-lib/spiutils/src/protocol/firmware.rs
```diff
@@ -262,9 +259,6 @@ impl ToWire for UpdatePrepareRequest {
 wire_enum! {
     /// The result of an update prepare request.
     pub enum UpdatePrepareResult: u8 {
-        /// Unknown result type.
-        Unknown = 0xff,
-
         /// Success
         Success = 0x00,
```
dann21, line 263: Would it make more sense for Success=1 and Failure=2, to more easily catch uninitialized values?  Or is that impossible in rust?

shared-lib/spiutils/src/protocol/firmware.rs
```diff
@@ -371,9 +365,6 @@ impl ToWire for WriteChunkRequest<'_> {
 wire_enum! {
     /// The result of a write chunk request.
     pub enum WriteChunkResult: u8 {
-        /// Unknown result type.
-        Unknown = 0xff,
-
         /// Success
         Success = 0x00,
```
dann21, line 369: Would it make more sense for Success=1 and Failure=2, to more easily catch uninitialized values?  Or is that impossible in rust?

shared-lib/spiutils/src/protocol/payload.rs
```diff
@@ -74,9 +74,6 @@ pub fn compute_checksum(header: &Header, payload: &[u8]) -> u8 {
 wire_enum! {
     /// The content type.
     pub enum ContentType: u8 {
-        /// Unknown message type.
-        Unknown = 0xff,
-
         /// Error
         Error = 0x00,
```
dann21, line 78: Would it make more sense for Errpr=1, etc., to more easily catch uninitialized values?  Or is that impossible in rust?

shared-lib/spiutils/src/protocol/firmware.rs
osenft: Rust wouldn't allow uninitialized values. You could argue that byte "0" on the wire is worse than byte "1", but for that we have the CRC to catch garbage. I'd leave it as is.

shared-lib/spiutils/src/protocol/firmware.rs
osenft: See above.

shared-lib/spiutils/src/protocol/payload.rs
osenft: See above.

PR #253, opened 2021-05-15
otpilot: Implement Firmware update protocol.

osenft:


shared-lib/spiutils/src/compat/firmware.rs
```diff
@@ -27,12 +27,17 @@ use core::mem;
 
 // ----------------------------------------------------------------------------
 
+/// The location of the BuildInfo from the start of the firmware segment.
```
dann21: s/location/offset/ ?

userspace/otpilot/src/spi_processor.rs
```diff
@@ -159,30 +169,191 @@ impl<'a> SpiProcessor<'a> {
...
+            let mut tx_cursor = SpiutilsCursor::new(&mut SPI_TX_BUF[payload::HEADER_LEN..]);
+
+            let fw_header = firmware::Header {
+                content: M::TYPE
+            };
+            fw_header.to_wire(&mut tx_cursor)?;
+            response.to_wire(&mut tx_cursor)?;
             payload_len = u16::try_from(tx_cursor.consumed_len())
-                .map_err(|_| SpiProcessorError::FromWire(FromWireError::OutOfRange))?;
+                    .map_err(|_| SpiProcessorError::FromWire(FromWireError::OutOfRange))?;
```
dann21: Nit: weird indentation.

dann21 (approved):
LGTM

shared-lib/spiutils/src/compat/firmware.rs
osenft: Ha, ok. Fixed.

userspace/otpilot/src/spi_processor.rs
osenft: Oh yeah, how did that happen? Thanks!

I'll need to run `rustfmt` on `spiutils`, `otpilot` and the new kernel code that I've written to make sure it conforms. I should have done that before and we should do that going forward. I'll send a follow up PR with just formatting changes.

PR #255, opened 2021-05-18
otpilot: Don't exceed version string buffer.

osenft:


userspace/otpilot/src/main.rs
```diff
@@ -85,7 +85,7 @@ fn run() -> TockResult<()> {
     {
         let mut idx : usize = 0;
         for val in "v1.00".as_bytes() {
-            if idx > identity.version.len() { break; }
+            if idx >= identity.version.len() { break; }
             identity.version[idx] = *val;
             idx = idx + 1;
         }
```
dann21: I know I don't know rust, but this seems like a really weird way to write this loop, and I'm not all surprised that this bug was written as a result.
Is this really the right way to iterate through the bytes that make up a string?

dann21 (approved):
LGTM.

userspace/otpilot/src/main.rs
osenft: I only found this bug in a later version of this code where the "v1.00" is NOT hardcoded but derived from the GIT commit ID at build time. 

But yeah, that code isn't good. Please have a look at the rewrite.

userspace/otpilot/src/main.rs
```diff
@@ -82,21 +84,13 @@ fn run() -> TockResult<()> {
...
 
-    {
-        let mut idx : usize = 0;
-        for val in "v1.00".as_bytes() {
-            if idx > identity.version.len() { break; }
-            identity.version[idx] = *val;
-            idx = idx + 1;
-        }
-    }
+    let banner = "v1.00".as_bytes();
```
dann21: Maybe call this `banner_bytes` so as to be parallel with the similar code immediately below?

userspace/otpilot/src/main.rs
```diff
@@ -82,21 +84,13 @@ fn run() -> TockResult<()> {
...
+    let max_len = min(identity.version.len(), banner.len());
```
dann21: If banner is too long, this will silently drop the tail end.  Is that OK?

userspace/otpilot/src/main.rs
```diff
@@ -82,21 +84,13 @@ fn run() -> TockResult<()> {
...
+    identity.version[..max_len].copy_from_slice(&banner[..max_len]);
 
-    {
-        let dev_id_bytes = fuse::get().get_dev_id()?.to_be_bytes();
-        for idx in 0..dev_id_bytes.len() {
-            identity.device_id[idx] = dev_id_bytes[idx];
-        }
-    }
+    let dev_id_bytes = fuse::get().get_dev_id()?.to_be_bytes();
+    let max_len = min(identity.device_id.len(), dev_id_bytes.len());
```
dann21, line 95: If `device_id_bytes` is too long, this will silently drop the tail end.  Is that OK?

dann21 (approved):
LGTM, modulo a naming nit and a couple of questions about possibly silently dropping information.

userspace/otpilot/src/main.rs
osenft: Yeah, I'd say that's ok. We're limited by the size that Manticore gives us in those fields.

osenft: We could print out an error message and continue ...

userspace/otpilot/src/main.rs
osenft: On Haven, dev_id_bytes will always be 8 bytes long, so not risk here. But apart from that, I don't think there's much we could do. We could print out an error message and continue ...

userspace/otpilot/src/main.rs
osenft: Makes sense. Done.

userspace/otpilot/src/main.rs
dann21: If no risk, then it should be an assertion.  If `min()` is necessary, then at the very least we should print something if we truncate.

Silent failure is *evil*.

userspace/otpilot/src/main.rs
```diff
@@ -82,21 +84,13 @@ fn run() -> TockResult<()> {
...
-    {
-        let mut idx : usize = 0;
-        for val in "v1.00".as_bytes() {
-            if idx > identity.version.len() { break; }
-            identity.version[idx] = *val;
-            idx = idx + 1;
-        }
-    }
+    let banner_bytes = "v1.00".as_bytes();
+    let max_len = min(identity.version.len(), banner_bytes.len());
```
dann21, line 88: If this `min()` call is really needed, then we should print an error message.  I'd be happy with an assertion here as well.  But silent truncation is a bad idea, IMESHO.

dann21 (approved):
Still quibbling about the spectre of silent truncation.

userspace/otpilot/src/main.rs
osenft: Fair. I added code to print an error.

userspace/otpilot/src/main.rs
osenft: Fair. I added code to print an error.

dann21 (approved):
LGTM

PR #256, opened 2021-05-24
Upgrade Tock to a more recent version and libtock-rs to HEAD

osenft:
This change upgrades Tock to just before Tock 2.0 Alpha1 and libtock-rs to current HEAD.

```
----------------------
`make prtest` summary:
----------------------
git rev-parse HEAD
b7b10190df57e4be4974be765b9b033816ca8a67
git status
On branch tock-libtock-rs-upgrade
Your branch is up to date with 'origin/tock-libtock-rs-upgrade'.

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,8 +99,8 @@ prtest: build check devicetests localtests
 # Installs the necessary Rust toolchains
 .PHONY: setup
 setup:
-	cd userspace && rustup target add thumbv7m-none-eabi && cd ..
-	cd kernel && rustup target add thumbv7m-none-eabi && cd ..
+	cd userspace && rustup target add thumbv7m-none-eabi
+	cd kernel && rustup target add thumbv7m-none-eabi
 
 
 # A target that prints an error message and fails the build if the cargo version
--- a/Makefile
+++ b/Makefile
@@ -99,14 +99,8 @@ prtest: build check devicetests localtests
 # Installs the necessary Rust toolchains
 .PHONY: setup
 setup:
-	rustup toolchain add --profile minimal \
-		"$$(cat third_party/libtock-rs/rust-toolchain)"
-	rustup toolchain add --profile minimal \
-		"$$(cat third_party/tock/rust-toolchain)"
-	rustup target add --toolchain \
-		"$$(cat third_party/libtock-rs/rust-toolchain)" thumbv7m-none-eabi
-	rustup target add --toolchain \
-		"$$(cat third_party/tock/rust-toolchain)" thumbv7m-none-eabi
+	cd userspace && rustup target add thumbv7m-none-eabi && cd ..
+	cd kernel && rustup target add thumbv7m-none-eabi && cd ..
 
 
 # A target that prints an error message and fails the build if the cargo version
--- a/kernel/h1/src/hil/flash/smart_program.rs
+++ b/kernel/h1/src/hil/flash/smart_program.rs
@@ -15,7 +15,7 @@
 /// A state machine for driving execution of the flash module's smart
 /// programming functionality.
 
-use ::kernel::hil::time::{Alarm,Frequency};
+use ::kernel::hil::time::{Alarm, Frequency};
 use ::kernel::ReturnCode;
 
 #[derive(Debug)]
@@ -55,7 +55,7 @@ impl SmartProgramState {
                 // Copied from Cr50: a timeout causes an immediate failure with
                 // no retry.
                 if hw.is_programming() {
-                    alarm.disable();
+                    alarm.disarm();
                     return Finished(ReturnCode::FAIL);
                 }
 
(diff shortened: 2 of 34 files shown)


userspace/test_harness/src/lib.rs
```diff
@@ -17,10 +17,13 @@
 // Relies on internal details of rustc, so this may break during toolchain
 // updates.
 
+#![feature(asm)]
```
jrvanwhy: Where is `feature(asm)` used in `test_harness`? Is it used by one of the new `libtock-rs` macros?

kernel/h1/src/timels.rs
```diff
@@ -74,46 +74,51 @@ impl Frequency for Freq256Khz {
...
+        self.now.get().wrapping_add(elapsed).into()
     }
 }
     
 impl Alarm<'static> for Timels {
-    fn set_alarm(&self, tics: u32) {
-        let distance = tics.wrapping_sub(self.now.get());
+    fn set_alarm(&self, _reference: Self::Ticks, dt: Self::Ticks) {
+        // TODO: This is over-simplified, as it does not account for time since
+        // reference.
```
jrvanwhy: IIRC you copied this TODO comment from my prototype, but you should address it. I think this logic needs to be fixed.

kernel/h1/src/timels.rs
```diff
@@ -74,46 +74,51 @@ impl Frequency for Freq256Khz {
...
 
-    fn disable(&self) {
+    fn disarm(&self) -> kernel::ReturnCode {
         let regs = unsafe { &*self.registers };
         regs.control.set(0);
+        kernel::ReturnCode::SUCCESS
+    }
+
+    fn minimum_dt(&self) -> Self::Ticks {
+        // TODO: placeholder value
```
jrvanwhy: This TODO should also be fixed. I _think_ you just need to search through the documentation and determine that `1` is correct.

Makefile
```diff
@@ -99,14 +99,8 @@ prtest: build check devicetests localtests
...
 setup:
-	rustup toolchain add --profile minimal \
-		"$$(cat third_party/libtock-rs/rust-toolchain)"
-	rustup toolchain add --profile minimal \
-		"$$(cat third_party/tock/rust-toolchain)"
-	rustup target add --toolchain \
-		"$$(cat third_party/libtock-rs/rust-toolchain)" thumbv7m-none-eabi
-	rustup target add --toolchain \
-		"$$(cat third_party/tock/rust-toolchain)" thumbv7m-none-eabi
+	cd userspace && rustup target add thumbv7m-none-eabi && cd ..
```
dann21: Better would be `(cd userspace && rustup target add thumbv7m-none-eabi)` -- that will execute the `cd` and the command in a sub-shell, so no matter what fails, you'll end up back in the directory where you started.  Same for below.  (Not sure if the parens are even necessary for Makefile commands, for which IIRC every command already gets executed in its own subshell, but not certain.)

kernel/h1/src/hil/flash/smart_program.rs
```diff
@@ -110,7 +110,7 @@ pub fn div_round_up(numerator: u64, denominator: u64) -> u64 {
 }
 
 fn set_program_timeout<'a, A: Alarm<'a>>(alarm: &A, timeout_nanoseconds: u32) {
-    alarm.set_alarm(alarm.now().wrapping_add(
-        div_round_up(A::Frequency::frequency() as u64 * timeout_nanoseconds as u64,
-                     1_000_000_000) as u32));
+    alarm.set_alarm(alarm.now(),
+        (div_round_up(A::Frequency::frequency() as u64 * timeout_nanoseconds as u64,
+                     1_000_000_000) as u32).into());
```
dann21, line 115: Nit: Weird indentation.  Maybe run `rustfmt` after, as a separate PR?

userspace/test_harness/src/lib.rs
osenft: Sorry about that, that got in there by mistake. Good catch. Removed.

kernel/h1/src/timels.rs
osenft: Oops, yeah. It ended up doing the right thing, but that was by chance due to how the capsule is implemented. I fixed it and retested.

kernel/h1/src/timels.rs
osenft: Yes, `1` is correct from what I can tell.

jrvanwhy (approved):

dann21 (commented):
LGTM, but I don't feel comfortable approving, given how little I understand about the surrounding ecosystem.

kernel/h1/src/hil/flash/smart_program.rs
osenft: Yes, I'll run rustfmt in a separate commit. Thanks for reminding me.

Makefile
osenft: From what I gathered each "command" (i.e. the whole && chain) is executed in a sub shell. So it's actually redundant to do a "cd .." at the end. I'll remove it.

dann21 (approved):
Aahhh -- what the hell, why not?

PR #257, opened 2021-05-25
Upgrade Manticore to current HEAD

osenft:


userspace/otpilot/src/manticore_support.rs
```diff
@@ -82,8 +89,8 @@ impl rsa::Engine for NoRsaEngine {
         &mut self,
         _signature: &[u8],
         _message: &[u8],
-    ) -> Result<(), ()> {
-        Err(())
+    ) -> Result<(), rsa::Error<()>> {
+        Err(manticore::crypto::rsa::Error::Custom(()))
```
dann21, line 90: When I was messing with this I used `Custom(())` as well.  Is that really right?  I would think there would be a way to yield a more specific indication of what went wrong.

dann21 (approved):
One question; LGTM otherwise.

userspace/otpilot/src/manticore_support.rs
osenft: I think it's ok for the time being, given that this is just "not implemented". We totally need to put something "real" here at some point.

PR #258, opened 2021-05-25
otpilot: Add support for Manticore Vendor firmware information.

osenft:


userspace/otpilot/src/main.rs
```diff
@@ -69,6 +72,20 @@ fn run_host_helper_demo() -> TockResult<()> {
     Ok(())
 }
 
+fn store_build_info(id: &str, segment_info: SegmentInfo, mut buf: &mut[u8]) {
```
dann21: Maybe add a comment that `id` is only used for error-spewage?

dann21 (approved):
LGTM

userspace/otpilot/src/main.rs
osenft: That parameter was actually redundant. I replaced it with code.

dann21 (approved):
LGTM

